PR #25, opened 2021-08-09
initial version of building tiledb core source

bdeng-xt:
add cmake options to build tiledb core for source. for example:
cmake -DTILEDB_VERSION=dev -DTILEDB_GIT_TAG=dev -DTILEDB_S3=ON -DTILEDB_AZURE=ON  ..

When TILEDB_VERSION is not in the format x.y.z, it will try to pull the branch TILEDB_GIT_TAG and build from it.


ihnorton:
Nice work, my only question is: will this run on ever CMake invocation?

bdeng-xt:
> Nice work, my only question is: will this build happen on every CMake invocation?
 The build will happen when we specify the TILEDB_VERSION which is not our release version format(x.y.z). I will update the default version from 2.3.0 to be 2.3..3 today

PR #27, opened 2021-08-10
change back vector initilization to Repeat method

bdeng-xt:
change back to use Repeat as initialization method

### DIFF
--- a/examples/ExampleDenseArray.cs
+++ b/examples/ExampleDenseArray.cs
@@ -123,8 +123,8 @@ namespace TileDB.Example
             subarray.Add(2);
             subarray.Add(4); //cols 2,3,4
 
-            TileDB.VectorInt32 data = new TileDB.VectorInt32(6); //hold 6 elements
-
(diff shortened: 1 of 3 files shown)


examples/ExampleDenseArray.cs
```diff
@@ -137,6 +137,10 @@ private static TileDB.QueryStatus ReadDenseSimpleArray()
             TileDB.QueryStatus status = query.submit();
             array.close();
 
+            for (int i = 0; i < data.Count; ++i)
+            {
+                System.Console.WriteLine("data at {0}:{1}", i, data[i]);
```
ihnorton, line 142: please add space after `:` here and below

ihnorton: also, printing the index `i` here doesn't really make sense because the array is 2D. so IMO we should drop that and just print the result vector.

ihnorton (approved):
Thanks. Please make the minor formatting change to output before merging.

examples/ExampleDenseArray.cs
bdeng-xt: changed to System.Console.WriteLine(String.Join(" ", data));

PR #28, opened 2021-08-11
Fix ch9553 ch9667 exception order and error message handling

ihnorton:
All we need is the exception type and error string, so remove the SWIG
class boilerplate for the exception classes. Now the error message is
passed through from libtiledb.

  before: `Exception of type 'TileDB.TypeError' was thrown.`

  after: `TileDB.TypeError: Static type (INT32) does not match expected
  type (UINT64)`

### DIFF
--- a/TileDB.CSharp/src/AttributeError.cs
+++ b/TileDB.CSharp/src/AttributeError.cs
@@ -11,35 +11,7 @@
 namespace TileDB {
 
 public class AttributeError : TileDBError {
-  private global::System.Runtime.InteropServices.HandleRef swigCPtr;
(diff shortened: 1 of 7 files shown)


bdeng-xt:
great. I was working the order of exception handling, so I just also pushed the changes to this branch. I changed the order in swig_tiledb_exception.i in the correct way(catch the derived error first, then TileDBError). so it will not have those build warnings.

bdeng-xt (approved):
I merged changes to this branch with another branch I was trying to fix the order of exception handling

ihnorton (commented):
💯Thanks! I forgot to un-stash the order part.

PR #42, opened 2021-10-31
Pack and UnPack string buffer

bdeng-xt:
This PR adds PackStringBuffer and UnPackStringBuffer functions in class TileDBBuffer. The end users don't need to handle offsets.

### DIFF
--- a/TileDB.CSharp/src/TileDBBuffer.cs
+++ b/TileDB.CSharp/src/TileDBBuffer.cs
@@ -34,50 +34,61 @@ namespace TileDB
         public TileDBBuffer()
         {
             m_elementType = typeof(T);
-            m_elementDataSize = (System.UInt32)(System.Runtime.InteropServices.Marshal.SizeOf(m_elementType));
+            
             if(m_elementType == typeof(System.Int32))
             {
                 m_elementDataType = TileDB.DataType.TILEDB_INT32;
+                m_elementDataSize = (System.UInt32)(System.Runtime.InteropServices.Marshal.SizeOf(m_elementType));
             }
             else if(m_elementType == typeof(System.Int64))
             {
                 m_elementDataType = TileDB.DataType.TILEDB_INT64;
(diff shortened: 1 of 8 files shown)


antalakas (approved):

TileDB.CSharp/src/TileDBBuffer.cs
```diff
@@ -123,11 +151,98 @@ protected void Release()
...
+            }
+            m_data = new T[strarray.Length];
+            m_offsets = TileDB.VectorUInt64.Repeat(0, strarray.Length);
+
+            List<byte[]> bytesList = new List<byte[]>();
+
+            int totsize = 0;
+            for(int i=0; i<strarray.Length; ++i)
+            {
+                byte[] bytes = Encoding.ASCII.GetBytes(strarray[i]);
```
ihnorton, line 177: It may be worth measuring, but I suspect we will be better off doing two passes here:
1. one pass to get the lengths using `Encoding.*.GetByteCount`
2. a single allocation for the output buffer
3. another pass to copy into the output buffer

(1 & 3 can be parallelized -- only if needed in the future for very large data)

ihnorton (commented):
Thanks, this is a great start. There's a couple things we need to do here:
- differentiate between ASCII and UTF8 (`Encoding.UTF8`)
- we should avoid making two copies of the data, see comments inline

TileDB.CSharp/src/TileDBBuffer.cs
bdeng-xt: Thanks for the suggestion. Added one parameter for encoding method, The default method is UTF8. We can also choose other method, such as ASCII, UTF16, and UTF32. As suggested, one of the two copies of the data is removed. Encoding.*.GetByteCount is used to get the total size of the buffer, then the encoded bytes are copied to the output buffer.

ihnorton (approved):

ihnorton:
👍

PR #43, opened 2021-11-19
support for file api reading

bdeng-xt:
This PR is an initial version for add reading support for file api. It added export_file_to_path helper function in ArrayUtil class


Shelnutt2 (approved):
Great first pass! Very fast addition thanks!

I'll leave it to you and @ihnorton on the follow-up items:
1) use vfs instead of ofstream for flexibility
2) take a parameter for size of buffer instead of allocating a vector for the full size up front

1) I only found a protected function xsputn from VFSFilebuf class to append bytes to a file, could you give me an example of using vfs to append
2) added a buffer_size as parameter

Shelnutt2:
Xref: @snagles

cpp/src/tiledb/cxx_api/tiledb_cxx_array_util.cc
```diff
@@ -9,33 +9,49 @@ std::string ArrayUtil::get_tiledb_version()
...
     uint64_t file_size = *((const int64_t*)v_r);
+    if(buffer_size==0) {
+      buffer_size = file_size;
+    }
+
+    std::string output_file = output_path;
+    if(output_file.empty()) {
+      tiledb::DataType file_name_datatype = tiledb::DataType::TILEDB_STRING_ASCII;
+      uint32_t name_size = 0;
+      char original_name[1024] = "";
```
Shelnutt2, line 35: You can leave this as a char*, it doesn't need any size. The pointer is set to the Array owned value.

cpp/src/tiledb/cxx_api/tiledb_cxx_array_util.cc
```diff
@@ -9,33 +9,49 @@ std::string ArrayUtil::get_tiledb_version()
...
+      array->get_metadata(METADATA_ORIGINAL_FILE_NAME,&file_name_datatype,&name_size,reinterpret_cast<const void**>(&original_name));
+      output_file = std::string(original_name);
```
Shelnutt2, line 37: This should be `output_file = std::string(original_name, name_size)`

cpp/src/tiledb/cxx_api/tiledb_cxx_array_util.cc
```diff
@@ -45,6 +61,10 @@ int ArrayUtil::export_file_to_path(const std::string& file_uri, const std::strin
         uint64_t read_size = result_buffer_elements[FILE_ATTRIBUTE_NAME][1];
         os.write((const char*)data_buffer.data(), read_size);
       }
+      offset += buffer_size;
+      subarray[0] = offset;
+      subarray[1] = (file_size > (offset + buffer_size)) ? (offset + buffer_size -1) : (file_size -1);
+      query.set_subarray(subarray);
```
Shelnutt2, line 65: You don't need to set the subarray each time, calling `submit` again will resume the query where it left off.

Shelnutt2 (changes requested):

bdeng-xt:
> Shouldn't this be `Query::Status::COMPLETE`?

The Status of Query is wrapped as QueryStatus

cpp/src/tiledb/cxx_api/tiledb_cxx_array_util.cc
bdeng-xt: Thanks for your info. I was not certain about that yesterday. Will correct it soon

ihnorton:
> The Status of Query is wrapped as QueryStatus

But we're using the C++ API here?

cpp/src/tiledb/cxx_api/tiledb_cxx_array_util.cc
bdeng-xt: corrected it. thanks

cpp/src/tiledb/cxx_api/tiledb_cxx_array_util.cc
bdeng-xt: corrected.

cpp/src/tiledb/cxx_api/tiledb_cxx_array_util.cc
bdeng-xt: removed those codes to set the subarray. thanks

cpp/src/tiledb/cxx_api/tiledb_cxx_array_util.cc
```diff
@@ -1,17 +1,165 @@
...
+      tiledb::DataType file_name_datatype = tiledb::DataType::TILEDB_STRING_ASCII;
+      uint32_t name_size = 0;
+      const char* original_name;
+      array->get_metadata(METADATA_ORIGINAL_FILE_NAME,&file_name_datatype,&name_size,reinterpret_cast<const void**>(&original_name));
+      output_file = std::string(original_name, name_size);
+    }
+    
+    std::shared_ptr<tiledb::VFS> vfs = std::shared_ptr<tiledb::VFS>(new tiledb::VFS(ctx));
+    tiledb::VFS::filebuf filebuf = tiledb::VFS::filebuf(vfs);
+    filebuf.open(output_file, std::ios::out);
```
ihnorton, line 46: Need to check for error here in case it is not writable

cpp/src/tiledb/cxx_api/tiledb_cxx_array_util.cc
```diff
@@ -1,17 +1,165 @@
...
+    std::ostream os(&filebuf);   
+    
+    uint64_t offset = 0;
+    std::array<uint64_t, 2> subarray = { offset, offset + buffer_size - 1 };
+
+    Query query(ctx, array, tiledb::QueryType::TILEDB_READ);
+    query.set_layout(tiledb::LayoutType::TILEDB_ROW_MAJOR)
+      .set_subarray(subarray);
+    std::vector<uint8_t> data_buffer;
```
ihnorton, line 55: This can be initialized with `buffer_size` as `data_buffer(buffer_size)`.

(In the future we should use a dedicated buffer class which does not initialize write zeros to the elements, because there's not need and it is expensive)

cpp/src/tiledb/cxx_api/tiledb_cxx_array_util.cc
```diff
@@ -1,17 +1,165 @@
...
+    data_buffer.resize(buffer_size);
+    query.set_buffer(FILE_ATTRIBUTE_NAME, data_buffer);
+
+    while (query.query_status() != QueryStatus::TILEDB_COMPLETED) {
```
ihnorton, line 59: This can give cause an infinite loop if the read never completes. We should have a counter here that checks the number of iterations without any results, and break out of the loop after something like 10 tries.

cpp/src/tiledb/cxx_api/tiledb_cxx_array_util.cc
```diff
@@ -1,17 +1,165 @@
...
+      query.submit();
+      auto result_buffer_elements = query.result_buffer_elements();
+      if (result_buffer_elements.find(FILE_ATTRIBUTE_NAME) != result_buffer_elements.end()) {
+        uint64_t read_size = result_buffer_elements[FILE_ATTRIBUTE_NAME][1];
```
ihnorton, line 63: eg
```
if (read_size == 0 && ++loop_zero_num > 10) { break; } else { read_size = 0; }
```

cpp/src/tiledb/cxx_api/tiledb_cxx_array_util.cc
```diff
@@ -1,17 +1,165 @@
...
+        os.write((const char*)data_buffer.data(), read_size);
+      }
+    }
 
+    array->close();
+    os.flush();
+    filebuf.close();
+  }
+  catch (const std::exception& e) {
```
ihnorton, line 72: We should catch `TileDBError` here and differentiate the error message.

cpp/src/tiledb/cxx_api/tiledb_cxx_array_util.cc
```diff
@@ -1,17 +1,165 @@
...
+    std::vector<uint8_t> contents;
+    contents.resize(nbytes);
+    std::array<uint64_t, 2> subarray = { 0,nbytes - 1 };
+    query.set_layout(tiledb::LayoutType::TILEDB_ROW_MAJOR);
+    query.set_subarray(subarray);
+    os.read((char*)contents.data(), nbytes);
+      
+    query.set_buffer(FILE_ATTRIBUTE_NAME, contents);
+    query.submit();
+    query.finalize();
```
ihnorton, line 138: We need to check the status here, because `finalize` resets the status.

cpp/src/tiledb/cxx_api/tiledb_cxx_array_util.cc
```diff
@@ -1,17 +1,165 @@
...
+    array->put_metadata(FILE_METADATA_MIME_TYPE_KEY, tiledb::DataType::TILEDB_STRING_ASCII,
+      static_cast<uint32_t>(mimetype.size()), mimetype.c_str());
+    std::string mimeencoding = mime_encoding.empty() ? "None" : mime_encoding;
+    array->put_metadata(FILE_METADATA_MIME_ENCODING_KEY, tiledb::DataType::TILEDB_STRING_ASCII,
+      static_cast<uint32_t>(mimeencoding.size()), mimeencoding.c_str());
 
+    array->close();
+    filebuf.close();
+  }
+  catch (const std::exception& e) {
```
ihnorton, line 154: Check `TileDBError`

ihnorton (approved):
Looking good, some comments inline. Also needs a test.

cpp/src/tiledb/cxx_api/tiledb_cxx_array_util.cc
Shelnutt2: Read queries should not call finalize

cpp/src/tiledb/cxx_api/tiledb_cxx_array_util.cc
bdeng-xt: added check if the return is nullptr or not.

cpp/src/tiledb/cxx_api/tiledb_cxx_array_util.cc
bdeng-xt: changed to data_buffer(buffer_size)

cpp/src/tiledb/cxx_api/tiledb_cxx_array_util.cc
bdeng-xt: added loop_zero_num as count variable

cpp/src/tiledb/cxx_api/tiledb_cxx_array_util.cc
bdeng-xt: changed, thanks

ihnorton: (updated to use correct variable)

cpp/src/tiledb/cxx_api/tiledb_cxx_array_util.cc
ihnorton: thanks

cpp/src/tiledb/cxx_api/tiledb_cxx_array_util.cc
```diff
@@ -1,17 +1,184 @@
...
+      if (result_buffer_elements.find(FILE_ATTRIBUTE_NAME) != result_buffer_elements.end()) {
+        read_size = result_buffer_elements[FILE_ATTRIBUTE_NAME][1];
+        os.write((const char*)data_buffer.data(), read_size);
+      }
 
+      if ( read_size == 0 && (++loop_zero_num) > 10) {
+        break;
+      }
+      else {
+        read_size = 0;
```
ihnorton, line 76: Sorry I made a mistake in my example, this should be `loop_zero_num = 0`

cpp/src/tiledb/cxx_api/tiledb_cxx_array_util.cc
bdeng-xt: added

cpp/src/tiledb/cxx_api/tiledb_cxx_array_util.cc
bdeng-xt: added

cpp/src/tiledb/cxx_api/tiledb_cxx_array_util.cc
bdeng-xt: Yes, this is from another function save_file_from_path. I tested with a very simple png image file(Sample.png, about 1Mb). First, use save_file_from_path to save the png file to a tiledb array., then use export_file_to_path to export from the tiledb array to another png file(Sample_exported.png). The size and content of the original file and exported file are the same.

ihnorton: We need to check `query_stats` after the write, _before_ calling `finalize`.

bdeng-xt:
> For @Shelnutt2's requests I think these are done?
> 
> * [x]  use vfs instead of ofstream for flexibility
We don't have vfs_file_handle in cpp-api. so I can only use VFS::VFSFilebuf to read and write. Also in VFS::VFSFilebuf class, xsgetn and xsputn are all protected functions. 

> * [x]  take a parameter for size of buffer instead of allocating a vector for the full size up front
already added a buffer_size parameter
 
> 
> Not sure about this -- @bdeng-xt please make sure to respond to questions as a separate comment, because no one is notified about edits.

cpp/src/tiledb/cxx_api/tiledb_cxx_array_util.cc
```diff
@@ -1,17 +1,184 @@
...
+      std::cout << "ArrayUtil::export_file_to_path, wrong buffer_size!" << std::endl;
+      return -1;
+    }
+
+    std::string output_file = output_path;
+    if(output_file.empty()) {
+      tiledb::DataType file_name_datatype = tiledb::DataType::TILEDB_STRING_ASCII;
+      uint32_t name_size = 0;
+      const char* original_name;
+      array->get_metadata(METADATA_ORIGINAL_FILE_NAME,&file_name_datatype,&name_size,reinterpret_cast<const void**>(&original_name));
```
ihnorton, line 40: Follow-up item: we need to check whether this is empty/unset, just in case.

cpp/src/tiledb/cxx_api/tiledb_cxx_array_util.cc
```diff
@@ -1,17 +1,184 @@
...
+    std::shared_ptr<tiledb::VFS> vfs = std::shared_ptr<tiledb::VFS>(new tiledb::VFS(ctx));
+    tiledb::VFS::filebuf filebuf = tiledb::VFS::filebuf(vfs);
+    if (filebuf.open(output_file, std::ios::out) == nullptr) {
+      std::cout << "ArrayUtil::export_file_to_path, can not open output_path:" << output_path << " for writing!" << std::endl;
+      return -1;
+    }
+    std::ostream os(&filebuf);   
+    
+    uint64_t offset = 0;
+    std::array<uint64_t, 2> subarray = { offset, offset + buffer_size - 1 };
```
ihnorton, line 53: Setting this subarray means we will only read at-most `buffer_size`, rather than the whole file -- we need to query for `file_size`. We handle reading in chunks in the `while` loop below.

ihnorton: I'm not sure what the goal is with `offset` here, seems unnecessary/cruft?

cpp/src/tiledb/cxx_api/tiledb_cxx_array_util.cc
bdeng-xt: added has_metadata check

cpp/src/tiledb/cxx_api/tiledb_cxx_array_util.cc
bdeng-xt: changed the subarray to {0,file_size-1} and removed offset

cpp/src/tiledb/cxx_api/tiledb_cxx_array_util.cc
ihnorton: We don't handle the empty and unavailable case, so if there is no file specified and no metadata key, we will probably try to write a file name of empty string.

cpp/src/tiledb/cxx_api/tiledb_cxx_array_util.cc
ihnorton: Bump.

cpp/src/tiledb/cxx_api/tiledb_cxx_array_util.cc
ihnorton: Bump

cpp/src/tiledb/cxx_api/tiledb_cxx_array_util.cc
ihnorton: Thanks

ihnorton:
I marked all of my issues except 2 resolved, not sure if @Shelnutt2 wants to re-review his comments.

cpp/src/tiledb/cxx_api/tiledb_cxx_array_util.cc
bdeng-xt: added check for empty file

cpp/src/tiledb/cxx_api/tiledb_cxx_array_util.cc
bdeng-xt: corrected. thanks

cpp/src/tiledb/cxx_api/tiledb_cxx_array_util.cc
bdeng-xt: Are we checking if query_status is INCOMPLETE and do a while loop to continue write?

ihnorton: No looping here, this is the write path, but we need to check `query_status == COMPLETE` or tell the user that their write did not complete.

cpp/src/tiledb/cxx_api/tiledb_cxx_array_util.cc
ihnorton: Sorry, what I mean is: we don't handle the condition where `output_file.empty() && !array->has_metadata(METADATA_ORIGINAL_FILE_NAME` -- no output file and no metadata filename available. IMO it should return an error, not try to write empty string or some undefined string.

Shelnutt2 (approved):

cpp/src/tiledb/cxx_api/tiledb_cxx_array_util.cc
bdeng-xt: thanks for the explanation.

cpp/src/tiledb/cxx_api/tiledb_cxx_array_util.cc
bdeng-xt: added the output when the query is not complete

cpp/src/tiledb/cxx_api/tiledb_cxx_array_util.cc
```diff
@@ -1,17 +1,195 @@
...
+    query.set_subarray(subarray);
+    os.read((char*)contents.data(), nbytes);
+      
+    query.set_buffer(FILE_ATTRIBUTE_NAME, contents);
+    tiledb::QueryStatus query_status = query.submit();
+    if (query_status == tiledb::QueryStatus::TILEDB_COMPLETED) {
+      query.finalize();
+    }
+    else {
+      std::cout << "ArrayUtil::save_file_from_path, query not completed!" << std::endl;
```
ihnorton, line 163: Shouldn't this be an error? We don't want them to think a file is written correctly if it is not actually complete... cc @Shelnutt2

Shelnutt2: Agreed but this API is only for testing, we do not expect end users to call the write functionality yet only read from C#, so we have flexibility.

bdeng-xt: added return statement when the write query is not complete

ihnorton: Thanks

examples/ExampleFileApi.cs
```diff
@@ -0,0 +1,102 @@
...
+                SaveFileToS3TileDBArray();
+            }
+
+            ExportS3TileDBArrayToFile();
+
+            return;
+        }
+
+        private static String array_uri_ = "test_file_array";
+        private static String s3_uri_ = "s3://stefan-region-test/image_tdb";
```
ihnorton, line 54: Please remove before merge.

ihnorton (approved):
Thanks for addressing all the comments, @bdeng-xt 💯

examples/ExampleFileApi.cs
ihnorton: Or make it a generic example uri.

bdeng-xt: removed, thanks

Shelnutt2:
Great work @bdeng-xt !

PR #46, opened 2021-11-22
Add nuspec file for nuget

bdeng-xt:
This PR prepared a nuspec file for creating nuget package. The directory structure is also adjusted to meet the requirement of nuget.  The build library for cpp is changed to TileDB.Cpp/lib. The create a nuget package, we can run the following command under TileDB.CSharp directory:
dotnet pack TileDB.CSharp.csproj /p:NuspecFile=TileDB.CSharp.nuspec -c Release


TileDB.CSharp/TileDB.CSharp.nuspec
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package >
+  <metadata>
+    <id>TileDB.CSharp</id>
+    <version>2.4.4</version>
```
Shelnutt2, line 5: Should we version the numbers separate from core?

bdeng-xt: It tries to make one-to-one correspondence between core and its wrapper. so the wrapper just use the same version number of core

TileDB.CSharp/TileDB.CSharp.csproj
```diff
@@ -14,39 +18,39 @@
...
+    <Content Include=".\lib\tiledbcs.dylib" CopyToOutputDirectory="PreserveNewest">
+      <Link>tiledbcs.dylib</Link>
     </Content>
   </ItemGroup>
-  <ItemGroup Condition=" '$([System.Runtime.InteropServices.RuntimeInformation]::IsOSPlatform($([System.Runtime.InteropServices.OSPlatform]::Linux)))' And Exists('..\dist\Linux\lib\libtiledb.so')">
+  <ItemGroup Condition=" '$([System.Runtime.InteropServices.RuntimeInformation]::IsOSPlatform($([System.Runtime.InteropServices.OSPlatform]::Linux)))'">
      <!-- this item is conditional in case we have a system-wide install -->
-     <Content Include="..\dist\Linux\lib\libtiledb.so" CopyToOutputDirectory="PreserveNewest">
-      <Link>tiledb.so</Link>
+     <Content Include=".\lib\libtiledb.so.2.4" CopyToOutputDirectory="PreserveNewest">
```
ihnorton, line 48: Do we need the unversioned so here? or are we directly linking `libtiledb.so.2.4`?

PR #54, opened 2021-12-25
Bd/ch12260 add metadata support

bdeng-xt:
add metadata support


cpp/src/tiledb/cxx_api/tiledb_cxx_array_util.cc
```diff
@@ -9,6 +9,127 @@ std::string ArrayUtil::get_tiledb_version()
...
+    result = "";
+  }
+  catch (...) {
+    std::cout << "ArrayUtil::get_array_schema_json_str, caught unknown error!" << std::endl;
+    result = "";
+  } 
+  return result;
+}//std::string ArrayUtil::get_array_schema_json_str(const std::string& uri) 
+    
+std::string ArrayUtil::get_array_metadata_json_str(const std::string& uri, const std::shared_ptr<tiledb::Context>& ctx) {
```
ihnorton, line 33: TileDB convention is that the context comes first.

cpp/src/tiledb/cxx_api/tiledb_cxx_array.h
```diff
@@ -1972,6 +2312,184 @@ class Array {
...
+        array_.get(),
+        index,
+        &key_c,
+        &key_len,
+        &value_type,
+        &value_num,
+        &value));
+
+    key.resize(key_len);
+    std::memcpy((void*)key.data(), key_c, key_len);
```
ihnorton, line 2338: Use the `std::string` constructor available for pointer + size.

TileDB.CSharp/src/CoreUtil.cs
```diff
@@ -0,0 +1,439 @@
...
+// TileDB Core lib helper functions
+namespace TileDB {
+
+    public class CoreUtil {
+
+        public static string GetCoreLibVersion() {
+            return ArrayUtil.get_tiledb_version();
+        }
+
+        public static TileDB.Array OpenArray(string uri, TileDB.QueryType querytype = TileDB.QueryType.TILEDB_READ, Context ctx = null)
```
ihnorton, line 11: This should be a static method on `TileDB.Array.

cpp/src/tiledb/cxx_api/tiledb_cxx_array_util.cc
```diff
@@ -9,6 +9,127 @@ std::string ArrayUtil::get_tiledb_version()
   return std::to_string(t[0]) + "." + std::to_string(t[1]) + "." + std::to_string(t[2]);
 }//std::string ArrayUtil::get_tiledb_version()
 
+std::string ArrayUtil::get_array_schema_json_str(const std::string& uri, const std::shared_ptr<tiledb::Context>& ctx) {
```
ihnorton, line 12: This should take a `TileDB.Array` object, not re-open the array.

ihnorton (commented):
- Please do not check in binaries.
- Adding vendored dependencies requires discussion and approval.

PR #55, opened 2021-12-27
change context to be the first parameter

bdeng-xt:
This PR is an attempt to addressing some problems in the PR#54(https://github.com/TileDB-Inc/TileDB-CSharp/pull/54). 
1.The Context parameter is moved to be the first parameter instead of the last. 
2. Binaries are removed from repo. At the mean time, I will manually copy those binaries to the same machine for packing. I will find a way to automatically to do that
3. Those static method in Util classes are just some helper functions for convenience. In stead of changing parameter uri to array object, I added extension methods to the Array class.


TileDB.CSharp/src/ArrayExtension.cs
```diff
@@ -0,0 +1,360 @@
...
+                System.Console.WriteLine(tdbe.Message);
+            }
+            catch (System.Exception e)
+            {
+                System.Console.WriteLine("caught exception:");
+                System.Console.WriteLine(e.Message);
+            }
+        }
+
+        public static void AddArrayMetadataByList<T>(this TileDB.Array array, string key, System.Collections.Generic.List<T> list) where T : struct
```
ihnorton, line 265: TileDB supports typed metadata, including multi-values for primitive types, so this seems to be redundant. Am I missing something here? IIUC this is encoding in JSON, storing the string, then decoding again into primitive types again on read. Is there a reason not to put the values as natively typed metadata?

bdeng-xt: Mostly for convenience reason. since metadata and schemadata are usually small, that will not be a performance issue. I will add another implementation in native typed data reading/writing, but would like to keep the json representaion of metadata and array schema.

ihnorton: The reason we have native metadata support is so that other API clients can read it. A custom JSON encoding is not compatible with other clients (we do use JSON for some python-specific extended type information in tiledb-py, but that is for attribute interpretation purposes; all other metadata is encoded natively).

bdeng-xt: Json is only an intermediate format to transfer data from c# to c++.  the final format is the same. Other api clients can read it without problem

ihnorton (approved):

cpp/CMakeLists.txt
```diff
@@ -62,7 +62,9 @@ if(BUILD_TYPE STREQUAL "DEBUG")
   add_compile_options(-DDEBUG )
 endif()
 
- 
+## relative path
+set(CMAKE_BUILD_RPATH "$\{ORIGIN\}")
```
ihnorton, line 66: Should already be handled here, I think: https://github.com/TileDB-Inc/TileDB-CSharp/blob/main/cpp/src/CMakeLists.txt#L114

bdeng-xt: then maybe we have to copy those two lib files to usr/lib, as suggested in https://github.com/dotnet/runtime/issues/34711

PR #59, opened 2022-01-13
Replace 2.6.0rc1 with 2.6.0 final

ihnorton:


ihnorton:
Why are we using sha1 here?

bdeng-xt (commented):
I changed TILEDB_DOWNLOAD_SHA1 to SHA1 value from SHA256. If there is more benefit to use SHA256, I can change all SHA1 to SHA256

ihnorton:
Yes, we should prefer sha256 wherever possible, thanks.

ihnorton:
For now, please just update 2.5 and use my update for 2.6

ihnorton:
No need to update all the others right now

PR #63, opened 2022-05-31
Multiplatform packaging

antalakas:


sources/TileDB.Interop/LibDllImport.cs
```diff
@@ -2,16 +2,13 @@
...
-        public const string LibCPath = "libc.dylib";
-#elif __LINUX__
-        public const string Path = "libtiledb.so.2.9";
-        public const string LibCPath = "libc.so";
-#else
-        public const string Path = "tiledb.dll";
-        public const string LibCPath = "msvcrt.dll";
-#endif
-   
+        public const string Path = "tiledb";
```
ihnorton, line 5: How about keeping the conditional `libtilebd.so.2.9` on linux and returning this on windows/mac?

antalakas: I think it will work, but what is the motivation for this? I guess it will make things a little bit easier with CI (only for the release) and the several xml files, but will leave us with the conditional for the OS which Imo is also very frustrating (with no reason as it turns out.) I am good with either case

ihnorton (approved):

.github/workflows/tiledb-csharp.yml
```diff
@@ -151,6 +151,13 @@ jobs:
     - name: Download tiledb
       run: ./.github/scripts/download_tiledb.sh
 
+    # Download tiledb
+    - name: Copy tiledb lib
+      shell: pwsh
+      run: |
+        $ErrorView = "NormalView"
+        cp sources/TileDB.CSharp/runtimes/linux-x64/native/libtiledb.so.2.9 sources/TileDB.CSharp/runtimes/linux-x64/native/libtiledb.so
```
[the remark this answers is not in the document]
antalakas: This just prepares in the CI container the file to have the proper naming for the resolver once the package gets installed. Only `libtiledb.so` will get in the package, so it will not be duplicated. This also means we need the physical file not a symlink

PR #64, opened 2022-06-09
Support filestore API

antalakas:


Shelnutt2 (approved):

sources/TileDB.CSharp/File.cs
```diff
@@ -1,172 +1,119 @@
...
+using System.Linq;
+using System.Reflection;
+using System.Runtime.InteropServices;
+using System.Text;
+using TileDB.Interop;
 
 namespace TileDB.CSharp
 {
-    public class File
+    public sealed unsafe class File
```
ihnorton, line 11: Since the API is still early, suggest to consider harmonizing the name to `Filestore` for consistency with core?

sources/TileDB.CSharp/File.cs
```diff
@@ -1,172 +1,119 @@
...
+        {
+            var ms_fileURI = new MarshaledString(fileURI);
+            var ms_arrayURI = new MarshaledString(arrayURI);
+            _ctx.handle_error(Methods.tiledb_filestore_uri_export(_ctx.Handle, ms_fileURI, ms_arrayURI));
+        }
+
+        public void BufferImport<T>(string arrayURI, T value, ulong size, MIMEType mimeType) where T: struct
+        {
+            var ms_arrayURI = new MarshaledString(arrayURI);
+            var data = new[] { value };
```
ihnorton, line 50: Isn't this going to make a copy of the value?

PR #66, opened 2022-07-01
Update to TileDB 2.10.1

antalakas:


cpp/cmake/TileDB.cmake
```diff
@@ -9,7 +9,18 @@ message(STATUS "start to set tiledb for version:${TILEDB_VERSION}")
 if(${TILEDB_VERSION} MATCHES "^([0-9]+)\\.([0-9]+)\\.([0-9]+)$")
 message(STATUS "start to set TILEDB_DOWNLOAD_URL and TILEDB_DOWNLOAD_SHA1")
 
-if(${TILEDB_VERSION} STREQUAL "2.9.4")
+if(${TILEDB_VERSION} STREQUAL "2.10.1")
```
Shelnutt2, line 12: Why do we do matching on the version?

examples/TileDB.CSharp.Example/TileDB.CSharp.Example.csproj
```diff
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <RootNamespace>TileDB.CSharp.Examples</RootNamespace>
-    <Version>5.0.3</Version>
+    <Version>5.0.4</Version>
```
Shelnutt2, line 5: When we change TileDB minor version we must always change minor version of the API. This needs to be 5.1. This is what we do with TileDB-Cloud-REST, TileDB-Go and all other APIs.

sources/TileDB.CSharp/TileDB.CSharp.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <RootNamespace>TileDB.CSharp</RootNamespace>
-    <Version>5.0.3</Version>
+    <Version>5.0.4</Version>
```
Shelnutt2, line 5: Bump minor version, should be 5.1.

sources/TileDB.CSharp/TileDB.CSharp.nuspec
```diff
@@ -2,14 +2,14 @@
 <package >
   <metadata>
     <id>TileDB.CSharp</id>
-    <version>5.0.3</version>
+    <version>5.0.4</version>
```
Shelnutt2, line 5: Bump minor version, should be 5.1.

sources/TileDB.Interop/TileDB.Interop.csproj
```diff
@@ -4,7 +4,7 @@
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
   </PropertyGroup>
   <PropertyGroup>
-    <Version>5.0.3</Version>
+    <Version>5.0.4</Version>
```
Shelnutt2, line 7: Bump minor version, should be 5.1.

Shelnutt2 (changes requested):

cpp/cmake/TileDB.cmake
ihnorton: This was carried over from the earlier API, there's definitions for the download targets for several versions below.

shaunrd0 (approved):

ihnorton (approved):

Shelnutt2 (approved):

PR #67, opened 2022-07-07
Add C# bindings quickstart_sparse_string example

shaunrd0:
TYPE: IMPROVEMENT
DESC: Add  C# bindings quickstart_sparse_string example


examples/bindings/quickstart_sparse_string/quickstart_sparse_string.csproj
```diff
@@ -0,0 +1,28 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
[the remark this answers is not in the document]
shaunrd0: I can definitely loop this into the current `TileDB.CSharp.Example` project. Alternatively I could create a new `TileDB.Interop.Example` project to group this and future Interop examples if we wanted to keep Interop and CSharp more separate.

I think for C# the current organization of `TileDB.CSharp.Example` makes sense and reduces a lot of repeated project scaffolding. This won't take too long so I'll go ahead and make changes now to add this example to `TileDB.CSharp.Example` and if we decide to go another route I can modify them again.

examples/bindings/quickstart_sparse_string/Program.cs
```diff
@@ -0,0 +1,321 @@
...
+            read_data_size_handle = GCHandle.Alloc(read_cols_size, GCHandleType.Pinned);
+            Methods.tiledb_query_set_data_buffer(ctx, read_query, cols_name,
+                read_data_handle.AddrOfPinnedObject().ToPointer(), (ulong*)read_data_size_handle.AddrOfPinnedObject()
+            );
+            read_data_handle.Free();
+            read_data_size_handle.Free();
+            check_err(ctx);
+
+            Methods.tiledb_query_submit(ctx, read_query);
+            check_err(ctx);
```
ihnorton, line 258: Check for query status here: it should be COMPLETE

examples/bindings/quickstart_sparse_string/Program.cs
```diff
@@ -0,0 +1,321 @@
...
+            Methods.tiledb_query_add_range(ctx, read_query, 1, &cols_start, &cols_end, null);
+            check_err(ctx);
+
+            Methods.tiledb_query_set_layout(ctx, read_query, tiledb_layout_t.TILEDB_ROW_MAJOR);
+            check_err(ctx);
+
+            var read_data_handle = GCHandle.Alloc(read_data, GCHandleType.Pinned);
+            var read_data_size_handle = GCHandle.Alloc(read_data_size, GCHandleType.Pinned);
+            Methods.tiledb_query_set_data_buffer(ctx, read_query, attr_name,
+                read_data_handle.AddrOfPinnedObject().ToPointer(), (ulong*)read_data_size_handle.AddrOfPinnedObject()
```
ihnorton, line 224: We need to use distinct variables/pointers for each of the buffer size handles here -- explanation below.

examples/bindings/quickstart_sparse_string/Program.cs
```diff
@@ -0,0 +1,321 @@
...
+            read_data_size_handle.Free();
+            check_err(ctx);
+
+            Methods.tiledb_query_submit(ctx, read_query);
+            check_err(ctx);
+
+            for (int i = 0; i < read_data.Length; i++)
+            {
+                // No more data to print if offset is 0 after first iteration
+                if (i > 0 && read_rows_offsets[i] == 0) break;
```
ihnorton, line 263: We should not rely on checking against 0 here, because this read might be out of bounds. Instead, when the `tiledb_query_submit` call returns above, the memory location pointed to by `read_data_size` (note: needs to be separate for each `set_..._buffer` call) will contain the total bytes read for that particular buffer. For offsets, we can divide by `sizeof(uint64_t)` to find the number of cells read. See the `buffer_size` doc comment here: https://tiledb-inc-tiledb.readthedocs-hosted.com/en/stable/c-api.html#c.tiledb_query_set_buffer

Finally, note that the size of the last cell must be calculated by subtracting the last offset from the total array size.

ihnorton (changes requested):
Some minor nits around result sizes handling. Overall this looks great!

examples/bindings/quickstart_sparse_string/Program.cs
```diff
@@ -0,0 +1,330 @@
...
+            data_size_handle.Free();
+            CheckError(ctx);
+
+            data_handle = GCHandle.Alloc(data_rows_offsets, GCHandleType.Pinned);
+            data_size_handle = GCHandle.Alloc(data_rows_offsets_size, GCHandleType.Pinned);
+            Methods.tiledb_query_set_offsets_buffer(ctx, query, rows_name,
+                (ulong*)data_handle.AddrOfPinnedObject(), (ulong*)data_size_handle.AddrOfPinnedObject()
+            );
+            data_handle.Free();
+            data_size_handle.Free();
```
ihnorton, line 156: Unless I am missing something about how GCHandle works, I believe these handles need to be kept alive until after the `tiledb_query_submit` call -- the size pointers are held internally until after query_submit returns.

ihnorton (approved):
LGTM, thanks for the updates! (please do a squash merge here, btw)

PR #69, opened 2022-07-18
 Add support for `TILEDB_BOOL`

shaunrd0:
Also added tests for bool and char attribute types. Found that `attribute.FillValue<char>();` returns an empty character array, created an exception to request use of `attribute.FillValue<byte>();` instead.


ihnorton (changes requested):
Thanks - could you please also add a test which reads and writes `bool` data?

antalakas (approved):

tests/TileDB.CSharp.Test/QueryTest.cs
```diff
@@ -400,5 +400,85 @@ public void TestNullableAttributeArrayQuery()
...
+
+            var a1 = new Attribute(context, "a1", DataType.TILEDB_BOOL);
+
+            var array_schema = new ArraySchema(context, ArrayType.TILEDB_DENSE);
+            array_schema.AddAttribute(a1);
+            array_schema.SetDomain(domain);
+            array_schema.Check();
+
+            var tmpArrayPath = Path.Join( Path.GetTempPath(), "tiledb_test_bool_array");
+            if (Directory.Exists(tmpArrayPath))
```
ihnorton, line 423: Side note - please make a story to create a fixture for this. In the future we will want to be able to run against non-local URIs and ensure, so centralizing temp directory logic will simplify things a lot.

tests/TileDB.CSharp.Test/QueryTest.cs
```diff
@@ -400,5 +400,85 @@ public void TestNullableAttributeArrayQuery()
...
+            query_read.SetSubarray(new [] {1, 2, 1, 2});
+            query_read.SetLayout(LayoutType.TILEDB_ROW_MAJOR);
+
+            var a1_data_read_bytes = new byte[4];
+            query_read.SetDataBuffer<byte>("a1", a1_data_read_bytes);
+
+            query_read.Submit();
+            status = query_read.Status();
+            Assert.AreEqual(status, QueryStatus.TILEDB_COMPLETED);
+            CollectionAssert.AreEqual(a1_data, System.Array.ConvertAll(a1_data_read_bytes, b => b == 1 ? true : false));
```
ihnorton, line 477: Please add a story for creating a higher-level interface to native types like this -- it would be very nice if users could get an array of `bool` back directly (like in Python or R) but we need to give flexibility to avoid the conversion cost too. Let's discuss at the next C# sync.

ihnorton (approved):
💯

tests/TileDB.CSharp.Test/QueryTest.cs
shaunrd0: [SC-19757](https://app.shortcut.com/tiledb-inc/story/19757/create-interface-for-c-native-types)

PR #70, opened 2022-07-18
Add CI for triggering releases

shaunrd0:
With these CI changes a TileDB-CSharp release can be triggered by pushing a new tag. After passing initial tests and building an updated NuGet package for this release, the workflow will upload `TileDB.CSharp.*.nupkg` as an artifact and later jobs will download this artifact to test the NuGet package locally on Windows, Mac, and Linux.

Any push to `main` will result in this workflow being triggered, but the final `Release` job will not run without a release tag. Always open to thoughts or suggestions on how this could be improved. 

Triggered this workflow manually (with no release tag): https://github.com/TileDB-Inc/TileDB-CSharp/actions/runs/2693214014

### DIFF
--- a/.github/workflows/tiledb-csharp.yml
+++ b/.github/workflows/tiledb-csharp.yml
@@ -2,7 +2,7 @@ name: TileDB-CSharp
 
 on:
   push:
-    tags: [ v* ]
+    tags: [ '*' ]
     branches: [ main ]
   pull_request:
     branches: [ main ]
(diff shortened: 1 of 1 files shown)


.github/workflows/tiledb-csharp.yml
```diff
@@ -2,172 +2,154 @@ name: TileDB-CSharp
...
   workflow_dispatch:
 
 jobs:
-  Linux_Test:
-    runs-on: ubuntu-latest
+  Run-Tests:
     strategy:
       matrix:
         # Will be checking following versions
         dotnet: ['5.0']
```
ihnorton, line 16: Please make an issue to eventually figure out how to run this against both 5.0 and 6.0 -- tried to do that at some point and got stuck with how to get .csproj supported versions spec to accept both.

.github/workflows/tiledb-csharp.yml
```diff
@@ -2,172 +2,154 @@ name: TileDB-CSharp
...
+        # Repeat this test for each os
+        os: [ubuntu-latest, macos-latest, windows-latest]
```
ihnorton, line 18: 👍

ihnorton (approved):
Looks great overall, nice work figuring out github actions 👏

PR #81, opened 2022-08-01
Add CI for .NET 6.0

shaunrd0:
Added `RollForward` tag to msbuild configurations to use .NET 6 if available. If both .NET 5.0 and .NET 6.0 SDKs are available and we use `dotnet run` without an explicit `--framework net5.0` .NET 6.0 will be used. The `--framework` flag should only be set if .NET 5.0 and 6.0 SDKs are both available and we want to target .NET 5.0. 

CI now tests building with both .NET 5.0 and 6.0 using new msbuild configurations. I used docker to test building in environments limited to either .NET 5.0 *or* .NET 6.0.

```bash
# From root of TileDB-CSharp/
docker run -v $(pwd):/app -w /app/examples/TileDB.CSharp.Example mcr.microsoft.com/dotnet/sdk:6.0 dotnet run
docker run -v $(pwd):/app -w /app/examples/TileDB.CSharp.Example mcr.microsoft.com/dotnet/sdk:5.0 dotnet run
```

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -25,7 +25,7 @@ jobs:
             HOST: win
             BOOTSTRAP: ../bootstrap.ps1 -EnableS3 -EnableSerialization
         tag: [release-2.10, release-2.11, dev]
-        dotnet: ['5.0']
+        dotnet: ['5.0', '6.0']
     runs-on: ${{ matrix.os }}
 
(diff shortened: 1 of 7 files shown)


ihnorton (approved):
Thanks for figuring this out

antalakas (approved):
Great work!

PR #84, opened 2022-08-10
Update to TileDB 2.11.0

shaunrd0:


.github/workflows/tiledb-csharp.yml
```diff
@@ -81,10 +81,6 @@ jobs:
       - name: Download tiledb
         run: ./.github/scripts/download_tiledb.sh
 
-      - name: Copy tiledb lib
```
Shelnutt2, line 84: Is this intentional to remove the copy?

Shelnutt2 (approved):
One question, beside that lgtm

.github/workflows/tiledb-csharp.yml
shaunrd0: Yep, `.github/scrips/dowload-tiledb.sh` grabs a release which for linux includes both `libtiledb.so.2.X` and `libtiledb.so`. [Here](https://github.com/TileDB-Inc/TileDB-CSharp/pull/84/files#diff-7d7d4853c64f077c7f69fe950ccc1eeb84d653ca36a3fccb8c65fce2ff416c9aL12) we extract the release to the `sources/TileDB.CSharp/runtimes/linux-x64/native/` directory so `libtiledb.so.2.X` is available without copying

PR #85, opened 2022-08-11
Support `TILEDB_FILTER_SCALE_FLOAT`

shaunrd0:
Also added support and test case for `TILEDB_FILTER_DICTIONARY`


Shelnutt2 (approved):

ihnorton (approved):
LGTM. Please add a story to write some tests actually exercising the filters -- this one can't round-trip exactly, but most of them are lossless and can be looped over one by one, checking that data round-trips correctly.

PR #94, opened 2022-08-24
Add ResultBufferElements to Query class

shaunrd0:
Also bumped minor version in preparation to release updated NuGet package


ihnorton (commented):
Belated, but LGTM 👍

PR #99, opened 2022-09-16
Add example and test for global order writes

shaunrd0:
Also normalized temp directory for tests and examples for [SC-19755](https://app.shortcut.com/tiledb-inc/story/19755/centralize-temp-directory-logic-in-mstests).

Renamed `quickstart_sparse_string` to `QuickstartSparseString`


sources/TileDB.CSharp/CoreUtil.cs
```diff
@@ -3,9 +3,23 @@
 using System;
 using System.Text;
 using System.Collections.Generic;
+using System.IO;
+
 namespace TileDB.CSharp {
 
-    public class CoreUtil {
+    public static class CoreUtil {
```
teo-tsirpanis, line 10: You should add this kind of code to a utilities file outside the main `TileDB.CSharp` library.

And making a class `static` is a breaking change because it does not allow constructing it and inheriting it (not that anybody would do these here but anyway).

teo-tsirpanis (changes requested):

sources/TileDB.CSharp/CoreUtil.cs
shaunrd0: Done :heavy_check_mark:

sources/TileDB.CSharp/CoreUtil.cs
```diff
@@ -3,10 +3,11 @@
 using System;
```
teo-tsirpanis, line 3: Please revert the changes in this file.

shaunrd0: done

ihnorton (approved):
LGTM, ping @teo-tsirpanis -- let's get it merged next week.

teo-tsirpanis (approved):

PR #101, opened 2022-10-03
Stop running tests in .NET 5 and retarget projects to .NET 6.

teo-tsirpanis:
.NET 5 has been unsupported since May 2022. This PR stops running CI in it, and retargets test and sample projects to .NET 6 instead.

I did not immediately retarget the main library projects but can do so as well.


ihnorton:
Thanks - we have to check with customers before we merge this.

teo-tsirpanis:
Closing, will revisit once we are able to move away from .NET 5.

PR #102, opened 2022-10-05
Add a NuGet package icon and set the `.nuspec` file as a project property.

teo-tsirpanis:


shaunrd0 (changes requested):
Thanks for doing this! :+1: The only nit I have is the size of the image should probably be the recommended NuGet size `128x128`. The current image is `460x460`

https://learn.microsoft.com/en-us/nuget/reference/nuspec#icon

teo-tsirpanis:
@shaunrd0 done.

shaunrd0 (approved):

PR #103, opened 2022-10-06
Use .NET's automatic library name resolution logic.

teo-tsirpanis:
As explained in https://learn.microsoft.com/en-us/dotnet/standard/native-interop/cross-platform#library-name-variations .NET has some automatic rules to allow loading native libraries in a cross-platform way. This PR takes advantage of this. In the P/Invoke definitions it changes the library name from `libtiledb` to `tiledb`, and removes the custom native library resolver; .NET will do the exact same thing.

I also removed commented-out mentions of `libc`; should we ever need to call `malloc`/`free` we should do it through the [`NativeMemory`](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.nativememory) class.


ihnorton (approved):
Nice! 👍 

From the docs, it looks like SO versioning is possible on linux. Do you know if/how it handles compatibility versioning on macOS?

teo-tsirpanis:
[Searching for `dylib_compatibility_version` in dotnet/runtime produced no results.](https://github.com/dotnet/runtime/search?q=dylib_compatibility_version) This concept seems very platform-specific for .NET to support.

PR #104, opened 2022-10-07
Fix `SonarAnalyzer.CSharp` not running.

teo-tsirpanis:
To be automatically imported by `MSBuild`, the property file has to be named `Directory.Build.props` instead of `Default.Build.props`. I fixed it, and also updated the analyzer to its latest version.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -2,9 +2,8 @@
   <ItemGroup>
     <PackageReference
       Include="SonarAnalyzer.CSharp"
-      Version="8.35.0.42613"
+      Version="8.46.0.54807"
       PrivateAssets="all"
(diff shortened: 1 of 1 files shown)


Directory.Build.props
```diff
@@ -2,9 +2,8 @@
   <ItemGroup>
     <PackageReference
       Include="SonarAnalyzer.CSharp"
-      Version="8.35.0.42613"
+      Version="8.46.0.54807"
       PrivateAssets="all"
-      Condition="$(MSBuildProjectExtension) == '.csproj'"
```
teo-tsirpanis, line 7: We don't need that; the analyzer will automatically not run on a language it does not recognize.

PR #105, opened 2022-10-07
Improve the bindings generation process and regenerate them.

teo-tsirpanis:
The bindings generation process has been streamlined in the following ways:

* `ClangSharpPInvokeGenerator` is installed as a local tool, ensuring we are in a common version. It was also updated to its latest one.
* Additional arguments have been passed to the generator such that we no longer need to manually edit `Methods.cs` afterwards. These arguments were also placed in `generate.rsp`.
* After these changes, the bindings were regenerated, with some breaking changes. I can't explain them, they also manifested with version 13.0.0-beta1 of the generator. But the previous versions were wrong. `size_t` maps to `nuint` in C# and not `ulong`, and nothing implied that enums are unsigned.

### DIFF
--- a/sources/TileDB.Interop/Methods.cs
+++ b/sources/TileDB.Interop/Methods.cs
@@ -1,13 +1,10 @@
+using System;
 using System.Runtime.InteropServices;
 
 namespace TileDB.Interop
 {
     public static unsafe partial class Methods
     {
-        [DllImport("tiledb", CallingConvention = CallingConvention.Cdecl, EntryPoint = "_Z13tiledb_statusi", ExactSpelling = true)]
-        [return: NativeTypeName("capi_status_t")]
-        public static extern int tiledb_status([NativeTypeName("capi_return_t")] int x);
-
         [DllImport("tiledb", CallingConvention = CallingConvention.Cdecl, ExactSpelling = true)]
         [return: NativeTypeName("int32_t")]
         public static extern int tiledb_query_type_to_str(tiledb_query_type_t query_type, [NativeTypeName("const char **")] sbyte** str);
@@ -106,6 +103,7 @@ namespace TileDB.Interop
 
         [DllImport("tiledb", CallingConvention = CallingConvention.Cdecl, ExactSpelling = true)]
         [return: NativeTypeName("const char *")]
+        [Obsolete]
         public static extern sbyte* tiledb_coords();
 
         [DllImport("tiledb", CallingConvention = CallingConvention.Cdecl, ExactSpelling = true)]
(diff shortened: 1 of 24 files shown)


teo-tsirpanis:
CI is being very flaky…

teo-tsirpanis:
The generator also learned to recognize `__declspec(deprecated)` and marked these APIs with `[Obsolete]`.

sources/TileDB.Interop/Methods.cs
```diff
@@ -1492,15 +1496,15 @@ public static unsafe partial class Methods
...
         [return: NativeTypeName("int32_t")]
-        public static extern int tiledb_filestore_buffer_import(tiledb_ctx_t* ctx, [NativeTypeName("const char *")] sbyte* filestore_array_uri, void* buf, [NativeTypeName("size_t")] ulong size, tiledb_mime_type_t mime_type);
+        public static extern int tiledb_filestore_buffer_import(tiledb_ctx_t* ctx, [NativeTypeName("const char *")] sbyte* filestore_array_uri, void* buf, [NativeTypeName("size_t")] nuint size, tiledb_mime_type_t mime_type);
 
         [DllImport("tiledb", CallingConvention = CallingConvention.Cdecl, ExactSpelling = true)]
         [return: NativeTypeName("int32_t")]
-        public static extern int tiledb_filestore_buffer_export(tiledb_ctx_t* ctx, [NativeTypeName("const char *")] sbyte* filestore_array_uri, [NativeTypeName("size_t")] ulong offset, void* buf, [NativeTypeName("size_t")] ulong size);
+        public static extern int tiledb_filestore_buffer_export(tiledb_ctx_t* ctx, [NativeTypeName("const char *")] sbyte* filestore_array_uri, [NativeTypeName("size_t")] nuint offset, void* buf, [NativeTypeName("size_t")] nuint size);
```
shaunrd0, line 1503: I ran into this `ulong` -> `nuint` conversion also when first regenerating the bindings. I have since been manually changing all of these back to `ulong` every time I regenerate, but I am very much in favor of getting this to a state where we won't have to manually edit these files after regenerating.

When I regenerate there are a lot more of these than what I see here, but I think it's likely due to updates that came with the ClangSharpPInvokeGenerator new 15.0.0 version. For me, all instances of `ulong` are replaced with `nuint`, but I notice for you it's only where the native type is `size_t`. 

We can sync on this during our 1:1 today to find out what's missing, but I think it's just that we are using different versions of ClangSharp - I will upgrade today and take another look here with the new version. Great work! :+1:

shaunrd0 (approved):
Looks great, thanks!

PR #106, opened 2022-10-13
Fix marshalling of `SafeHandle`s et. al.

teo-tsirpanis:
Up until now we have been underutilizing our safe handles by not taking advantage of their reference counting mechanism, making them prone to disposals by other threads in the middle of using them.

This PR fixes that, ensuring that the reference count is incremented and decremented before and after a PInvoke call. I also fixed some formatting issues, replaced `stackalloc` statements of size 1 with just declaring a variable, and making the safe handles internal and moving them to their own namespace inside the `TileDB.CSharp` project.

[SC-22342](https://app.shortcut.com/tiledb-inc/story/22342/the-reference-counting-mechanism-of-safe-handles-is-not-used)


sources/TileDB.CSharp/Marshalling/SafeHandles/ArrayHandle.cs
```diff
@@ -0,0 +1,53 @@
...
+            finally
+            {
+                if (successful)
+                {
+                    handle.InitHandle(array);
+                }
+            }
+
+            return handle;
+        }
```
teo-tsirpanis, line 37: We mimic the way the .NET runtime marshals safe handles from native code. Instead of just a `new ArrayHandle((IntPtr)array, ownsHandle: true)`, we have to first create it with an invalid value and then assign it before we return it.

image

sources/TileDB.CSharp/ArraySchema.cs
```diff
@@ -208,8 +233,10 @@ public FilterList CoordsFilterList()
         {
 
             tiledb_filter_list_t* filter_list_p;
-            _ctx.handle_error(Methods.tiledb_array_schema_get_coords_filter_list(_ctx.Handle, _handle, &filter_list_p));
-            return new FilterList(_ctx, filter_list_p);
+            using var ctxHandle = _ctx.Handle.Acquire();
+            using var handle = _handle.Acquire();
+            _ctx.handle_error(Methods.tiledb_array_schema_get_coords_filter_list(ctxHandle, handle, &filter_list_p));
+            return new FilterList(_ctx, FilterListHandle.CreateUnowned(filter_list_p));
```
teo-tsirpanis, line 239: An unowned handle means that if we call `Dispose` it will not be released from the native side. That's what the pre-existing implicit conversion operator did. Is that surely the correct behavior? If not I will fix it in a future PR.

teo-tsirpanis (commented):
@ihnorton, @shaunrd0 can you review it? It's a prerequisite for further PRs.

sources/TileDB.CSharp/Array.cs
```diff
@@ -398,11 +427,6 @@ public static void Vacuum(Context ctx, string uri, Config config)
                         nonEmptyDomain.Add(dimName, new Tuple<string, string>(data0, data1));
                     }
                         break;
-                    default:
-                        {
-                        }
```
shaunrd0, line 403: This might be a good place to throw?

sources/TileDB.CSharp/ArraySchema.cs
shaunrd0: It seems to me that this is unintended, but worth checking with @antalakas if there's a reason for this I may be unaware of.

sources/TileDB.CSharp/Group.cs
```diff
@@ -252,18 +258,18 @@ public ulong MemberCount()
         /// </summary>
         /// <param name="index"></param>
         /// <returns></returns>
-        (string uri, ObjectType object_type, string name) MemberByIndex(ulong index) 
+        (string uri, ObjectType object_type, string name) MemberByIndex(ulong index)
```
shaunrd0, line 261: Good catch on the typo! I think we could also make this method public and then add to `tests/TileDB.CSharp.Test/GroupTest.cs` -

```C#
//Reopen in read mode
group1.Open(QueryType.TILEDB_READ);
Assert.AreEqual<ulong>(2, group1.MemberCount());
Assert.AreEqual("array2", group1.MemberByIndex(0).name);
Assert.AreEqual("array1", group1.MemberByIndex(1).name);
group1.Close();
```

I did this locally using this branch and the test passes, just thinking while we're here it would be nice to test for this to protect against future regressions.

sources/TileDB.CSharp/Marshalling/SafeHandles/ArraySchemaEvolutionHandle.cs
```diff
@@ -0,0 +1,53 @@
...
+
+namespace TileDB.CSharp.Marshalling.SafeHandles
+{
+    internal unsafe sealed class ArraySchemaEvolutionHandle : SafeHandle
+    {
+        public ArraySchemaEvolutionHandle() : base(IntPtr.Zero, true) { }
+
+        public ArraySchemaEvolutionHandle(IntPtr handle, bool ownsHandle) : base(IntPtr.Zero, ownsHandle) { SetHandle(handle); }
+
+        public static ArraySchemaEvolutionHandle CreateUnowned(tiledb_array_t* array) => new((IntPtr)array, ownsHandle: false);
```
shaunrd0, line 13: Typo? `CreateUnowned(tiledb_array_t* array)` ->  `CreateUnowned(tiledb_array_schema_evolution_t* array)`

sources/TileDB.CSharp/Marshalling/SafeHandles/ContextHandle.cs
```diff
@@ -0,0 +1,75 @@
...
+
+namespace TileDB.CSharp.Marshalling.SafeHandles
+{
+    internal unsafe sealed class ContextHandle : SafeHandle
+    {
+        public ContextHandle() : base(IntPtr.Zero, true) { }
+
+        public ContextHandle(IntPtr handle, bool ownsHandle) : base(IntPtr.Zero, ownsHandle) { SetHandle(handle); }
+
+        public static ContextHandle CreateUnowned(tiledb_ctx_t* schema) => new((IntPtr)schema, ownsHandle: false);
```
shaunrd0, line 14: It's a variable name so just a nit, but I think `schema` should be `context` here. Similar typo for this variable on other safehandle classes.

shaunrd0 (changes requested):
A few small typos / suggestions, but no major concerns - this looks great. Thanks for the work you did here!

sources/TileDB.CSharp/Array.cs
teo-tsirpanis: I prefer to handle it separately.

sources/TileDB.CSharp/Group.cs
teo-tsirpanis: I prefer to handle it separately. The API signature also needs some improvements.

sources/TileDB.CSharp/Marshalling/SafeHandles/ArraySchemaEvolutionHandle.cs
teo-tsirpanis: Copy-pasting oversights. 😅 Fixed.

PR #118, opened 2022-10-20
Fix race condition in `Query.SubmitAsync()`.

teo-tsirpanis:
The implementation of `Query.SubmitAsync()` had a race condition. According to the documentation of [`Marshal.GetFunctionPointerForDelegate`](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.marshal.getfunctionpointerfordelegate) _"You must manually keep the delegate from being collected by the garbage collector from managed code. The garbage collector does not track references to unmanaged code."_. We did not that, and if the garbage collector collects the delegate in the middle of an asynchronous query, by the time the native code calls the callback, it will be gone, and (I think) the process will crash.

This PR fixes the problem in two ways. First we mark the C# callback function with [`[UnmanagedCallersOnly]`](https://devblogs.microsoft.com/dotnet/improvements-in-native-code-interop-in-net-5-0/#unmanagedcallersonly), allowing us to directly pass the function's pointer to native code without allocating a delegate. Second we ensure that the `Query` object remains alive during the asynchronous query, by allocating a `GCHandle`, and passing it as the `callback_data` parameter to `tiledb_query_submit_async`.

### DIFF
--- a/sources/TileDB.CSharp/Query.cs
+++ b/sources/TileDB.CSharp/Query.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 using System.Text;
 using TileDB.CSharp.Marshalling.SafeHandles;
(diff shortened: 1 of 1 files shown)


teo-tsirpanis:
Superseded by #120.

PR #120, opened 2022-10-21
Make `Query.SubmitAsync` return `ValueTask`.

teo-tsirpanis:
This PR changes the way queries are submitted asynchronously. `Query.SubmitAsync` was modified to return `ValueTask`, allowing us to use the Task Asynchronous Pattern and the `async`/`await` keywords.

Instead of

```csharp
query.QueryCompleted += OnCompleted;
query.SubmitAsync();

void OnCompleted(object sender, QueryEventArgs args)
{
    // continue
}
```

all we have to write is

```csharp
await query.SubmitAsync();
// continue
```
.

The usability benefits are enormous. Callback hell is avoided, and by making asynchronous queries as easy as synchronous ones, their usage can be popularized. There are also performance -the `ValueTask` is powered by [a reusable state machine object](https://blog.marcgravell.com/2019/08/prefer-valuetask-to-task-always-and.html) which means that calling `await query.SubmitAsync()` on the same query object many times will allocate only once- and correctness benefits -besides the bug #118 would fix which is fixed here as well, the previous implementation did not correctly propagate some ambient state to the callback-.

The old event-based code was entirely removed. This is a breaking change but one that I believe is definitely worth it.

The query tests were updated to use both `Submit` and `SubmitAsync`, and the examples were converted to exclusively use `SubmitAsync`.

[SC-22350](https://app.shortcut.com/tiledb-inc/story/22350/support-the-task-asynchronous-pattern-in-asynchronous-queries)

### DIFF
--- a/examples/TileDB.CSharp.Example/ExampleIncompleteQuery.cs
+++ b/examples/TileDB.CSharp.Example/ExampleIncompleteQuery.cs
@@ -69,14 +69,14 @@ namespace TileDB.CSharp.Examples
             }
         }
 
-        static void ReadArray()
+        static async Task ReadArrayAsync()
         {
             // Allocate buffers for 10 values per batch read
             var rowsRead = new int[10];
             var colsRead = new int[10];
             var attrRead = new int[10];
 
-            using (var arrayRead = new Array(ctx, arrayName))
+            using (var arrayRead = new Array(Ctx, ArrayPath))
             {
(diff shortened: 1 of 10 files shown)


teo-tsirpanis:
Let's close this, it cannot be merged until `tiledb_query_submit_async` gets fixed, which is going to take time.

PR #126, opened 2022-10-26
Update CI

### LINKED ISSUE #107: Nightly GitHub Actions Build Fail on Fri, October 14th 2022
See run for more details:
https://github.com/TileDB-Inc/TileDB-CSharp/actions/runs/3247204541

shaunrd0:
Bumps nightly builds to test against 2.11 and later versions. Used xargs to run examples since `find --exec-dir` [will not fail CI](https://github.com/TileDB-Inc/TileDB-CSharp/actions/runs/3228846664/jobs/5285461303#step:9:874) if an example produces errors. Updated nightly build issue template to assign to @teo-tsirpanis

Passing manual run for [nightly CI](https://github.com/TileDB-Inc/TileDB-CSharp/actions/runs/3330919709/jobs/5510034556). Closes #107

### DIFF
--- a/.github/workflows/nightly-issue-template.md
+++ b/.github/workflows/nightly-issue-template.md
@@ -1,6 +1,6 @@
 ---
 title: Nightly GitHub Actions Build Fail on {{ date | date('ddd, MMMM Do YYYY') }}
-assignees: shaunrd0
+assignees: teo-tsirpanis
 labels: bug
 ---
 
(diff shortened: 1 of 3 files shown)


teo-tsirpanis (approved):
LGTM. I wonder why we are testing older Core versions. If we add a feature that got introduced in a new Core version, and a test for it, how would the older version handle it?

shaunrd0:
> If we add a feature that got introduced in a new Core version, and a test for it, how would the older version handle it?

We ran into this [here](https://github.com/TileDB-Inc/TileDB-CSharp/issues/87). Initially I was thinking we would bump CI to test against newer versions, but after discussing during C# sync we decided instead to skip tests for features that were introduced in later versions, and this was merged with #91. I noticed the description for this PR was lacking explanation so I added a brief description to better explain its purpose.

PR #128, opened 2022-10-27
Add `FragmentInfo`.

teo-tsirpanis:
This PR implements TileDB's fragment info API, and adds tests to it, ported from Java.

[SC-22321](https://app.shortcut.com/tiledb-inc/story/22321/add-high-level-wrapper-for-fragmentinfo-api)

### DIFF
--- a/sources/TileDB.CSharp/Enums.cs
+++ b/sources/TileDB.CSharp/Enums.cs
@@ -1105,6 +1105,11 @@ namespace TileDB.CSharp
         public static tiledb_datatype_t to_tiledb_datatype(Type t) =>
             (tiledb_datatype_t)TypeToDataType(t);
 
+        /// <summary>
+        /// Converts a <see cref="Type"/> to a <see cref="DataType"/> enum value.
+        /// </summary>
+        /// <param name="t">The type to convert.</param>
+        /// <exception cref="NotSupportedException"><paramref name="t"/> is unsupported.</exception>
         public static DataType TypeToDataType(Type t)
         {
             if (t == typeof(int))
@@ -1157,15 +1162,10 @@ namespace TileDB.CSharp
             }
             else
             {
-#pragma warning disable CS0618 // Type or member is obsolete
-                return DataType.Any;
-#pragma warning restore CS0618 // Type or member is obsolete
+                ThrowHelpers.ThrowTypeNotSupported();
+                return default; // unreachable
             }
         }
-
-        [Obsolete("This method will be removed in a future version of the library. Use DataTypeToType and the DataType enum instead.")]
-        private static Type TileDBDataTypeToType(tiledb_datatype_t tiledbDatatype) =>
-            DataTypeToType((DataType)tiledbDatatype);
 
         public static Type DataTypeToType(DataType datatype)
         {
--- a/sources/TileDB.CSharp/Enums.cs
+++ b/sources/TileDB.CSharp/Enums.cs
@@ -282,12 +282,6 @@ namespace TileDB.CSharp
         // https://github.com/TileDB-Inc/TileDB/pull/2812
         StringUcs4 = tiledb_datatype_t.TILEDB_STRING_UCS4,
         /// <summary>
-        /// Any value. Deprecated.
-        /// </summary>
-        [Obsolete("This data type is deprecated.")]
-        // https://github.com/TileDB-Inc/TileDB/pull/2807
-        Any = tiledb_datatype_t.TILEDB_ANY,
-        /// <summary>
         /// A date and time, counted as the signed 64-bit number of
         /// years since the Unix epoch (January 1 1970 at midnight).
         /// </summary>
@@ -439,8 +433,8 @@ namespace TileDB.CSharp
         TILEDB_STRING_UCS2 = StringUcs2,
         [Obsolete("Use StringUcs4 instead."), EditorBrowsable(EditorBrowsableState.Never)]
         TILEDB_STRING_UCS4 = StringUcs4,
-        [Obsolete("Use Any instead."), EditorBrowsable(EditorBrowsableState.Never)]
-        TILEDB_ANY = Any,
+        [Obsolete("This data type is deprecated."), EditorBrowsable(EditorBrowsableState.Never)]
+        TILEDB_ANY = tiledb_datatype_t.TILEDB_ANY,
         [Obsolete("Use DateTimeYear instead."), EditorBrowsable(EditorBrowsableState.Never)]
         TILEDB_DATETIME_YEAR = DateTimeYear,
         [Obsolete("Use DateTimeMonth instead."), EditorBrowsable(EditorBrowsableState.Never)]
@@ -1172,8 +1166,6 @@ namespace TileDB.CSharp
 #pragma warning disable CS0618 // Type or member is obsolete
             switch (datatype)
             {
-                case DataType.Any:
-                    return typeof(sbyte);
                 case DataType.Char:
                     return typeof(sbyte);
                 case DataType.DateTimeAttosecond:
--- a/sources/TileDB.CSharp/FragmentInfo.cs
+++ b/sources/TileDB.CSharp/FragmentInfo.cs
@@ -411,12 +411,14 @@ namespace TileDB.CSharp
         /// with the dimension's data type.</exception>
         public (T Start, T End) GetMinimumBoundedRectangle<T>(uint fragmentIndex, uint minimumBoundedRectangleIndex, uint dimensionIndex)
         {
-            ValidateDomainType<T>(GetDimensionType(fragmentIndex, dimensionIndex));
+            DataType dataType = GetDimensionType(fragmentIndex, dimensionIndex);
+            ValidateDomainType<T>(dataType);
             if (RuntimeHelpers.IsReferenceOrContainsReferences<T>())
             {
                 if (typeof(T) == typeof(string))
                 {
-                    (string startStr, string endStr) = GetStringMinimumBoundedRectangle(fragmentIndex, minimumBoundedRectangleIndex, dimensionIndex);
+                    (string startStr, string endStr) =
+                        GetStringMinimumBoundedRectangle(fragmentIndex, minimumBoundedRectangleIndex, dimensionIndex, dataType);
                     return ((T)(object)startStr, (T)(object)endStr);
                 }
                 ThrowHelpers.ThrowTypeNotSupported();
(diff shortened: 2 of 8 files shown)


sources/TileDB.CSharp/Marshalling/ScratchBuffer.cs
```diff
@@ -0,0 +1,76 @@
...
+namespace TileDB.CSharp.Marshalling
+{
+    /// <summary>
+    /// Manages temporary buffers with low GC pressure.
+    /// </summary>
+    /// <remarks>
+    /// Objects of this type should not be passed around; instead pass their <see cref="Span"/> property.
+    /// </remarks>
+    /// <typeparam name="T">The type of items this scratch buffer stores.</typeparam>
+    internal ref struct ScratchBuffer<T>
```
teo-tsirpanis, line 16: Copied from https://gist.github.com/teo-tsirpanis/db9ae014efcad311a5df87e6ed04bc51. I can strip license headers, it's mine and in public domain.

tests/TileDB.CSharp.Test/FragmentInfoTest.cs
```diff
@@ -0,0 +1,511 @@
...
+            Assert.AreEqual(1u, info.FragmentCount);
+
+            Assert.AreEqual(2ul, info.GetMinimumBoundedRectangleCount(0));
+
+            Assert.AreEqual(("a", "bb"), info.GetMinimumBoundedRectangle<string>(0, 0, 0));
+            Assert.AreEqual(("c", "ddd"), info.GetMinimumBoundedRectangle<string>(0, 1, "d"));
+        }
+
+        [TestMethod]
+        public void TestMinimumBoundedRectangles()
```
ihnorton, line 177: Can we parametrize this over different numeric types?

teo-tsirpanis: I copy-pasted the MBR test to work with attributes with 32-bit integers. We could avoid duplication with [.NET 7's generic math](https://learn.microsoft.com/en-us/dotnet/standard/generics/math) but until then…

sources/TileDB.CSharp/FragmentInfo.cs
```diff
@@ -0,0 +1,660 @@
...
+        /// <param name="fragmentIndex">The index of the fragment of interest.</param>
+        /// <param name="minimumBoundedRectangleIndex">The index of the MBR of interest.</param>
+        /// <param name="dimensionIndex">The index of the dimension of interest, following
+        /// the order as it was defined in the domain of the array schema.</param>
+        /// <returns>The start and end values of the MBR, inclusive.</returns>
+        /// <exception cref="NotSupportedException"><typeparamref name="T"/> is a managed type
+        /// other than <see cref="string"/>.</exception>
+        /// <remarks>
+        /// In case of a type mismatch this function will not fail but return invalid data.
+        /// </remarks>
```
teo-tsirpanis, line 414: Thoughts about this? To check the generic type we would have to allocate an `ArraySchema`; the function currently does not allocate. Is it OK to return garbage if the user specifies the wrong type? This is an advanced API either way.

teo-tsirpanis: Done.

sources/TileDB.CSharp/Enums.cs
```diff
@@ -1157,16 +1162,11 @@ public static DataType TypeToDataType(Type t)
             }
             else
             {
-#pragma warning disable CS0618 // Type or member is obsolete
-                return DataType.Any;
-#pragma warning restore CS0618 // Type or member is obsolete
+                ThrowHelpers.ThrowTypeNotSupported();
+                return default; // unreachable
```
teo-tsirpanis, line 1166: Thoughts on this? Returning `Any` on an unsupported type can be misleading since `Any` is obsolete and there is no clear documentation on what to do with it.

sources/TileDB.CSharp/Marshalling/MarshaledStringOut.cs
```diff
@@ -11,6 +12,16 @@ internal static class MarshaledStringOut
...
             Encoding.ASCII.GetString(span);
 
+        public static string GetString(ReadOnlySpan<byte> span, DataType dataType) =>
+            dataType switch
+            {
+                DataType.StringAscii => Encoding.ASCII.GetString(span),
+                DataType.StringUtf8 => Encoding.UTF8.GetString(span),
+                DataType.StringUtf16 => Encoding.Unicode.GetString(span),
+                DataType.StringUtf32=> Encoding.UTF32.GetString(span),
+                _ => throw new ArgumentOutOfRangeException(nameof(dataType), dataType, "Unsupported string data type.")
```
teo-tsirpanis, line 22: I didn't handle UCS-2 and UCS-4; they are legacy, don't have framework support and need custom code to convert. Is that OK?

teo-tsirpanis:
[CI is displaying these errors.](https://github.com/TileDB-Inc/TileDB-CSharp/actions/runs/3523622203/jobs/5908294088#step:8:43)

sources/TileDB.CSharp/Marshalling/MarshaledStringOut.cs
ihnorton: Yes

sources/TileDB.CSharp/Enums.cs
ihnorton: What is `default` here?

ihnorton: I think throwing is fine, if you mean `TILEDB_ANY`, there's definitely no way to translate that.

teo-tsirpanis: In general `default` is equal to `null` on reference types, zero on primitives and all fields recursively set to `default` in value types.

In this case I added it to satisfy the compiler. It won't be returned because the above function will throw (we do it in separate method for performance reasons).

sources/TileDB.CSharp/FragmentInfo.cs
```diff
@@ -0,0 +1,681 @@
...
+        /// </summary>
+        /// <param name="encryptionType">The encryption type to use.</param>
+        /// <param name="key">The encryption key to use.</param>
+        public void LoadWithKey(EncryptionType encryptionType, ReadOnlySpan<byte> key)
+        {
+            using var ctxHandle = _ctx.Handle.Acquire();
+            using var handle = _handle.Acquire();
+            fixed (byte* keyPtr = key)
+            {
+                _ctx.handle_error(Methods.tiledb_fragment_info_load_with_key(ctxHandle, handle, (tiledb_encryption_type_t)encryptionType, keyPtr, (uint)key.Length));
```
teo-tsirpanis, line 60: This API is obsolete. The keys are now supposed to be set through the `Config`, right?

sources/TileDB.CSharp/Enums.cs
```diff
@@ -1105,6 +1099,11 @@ public static VfsMode VfsModeFromStr(string str)
         public static tiledb_datatype_t to_tiledb_datatype(Type t) =>
             (tiledb_datatype_t)TypeToDataType(t);
 
+        /// <summary>
+        /// Converts a <see cref="Type"/> to a <see cref="DataType"/> enum value.
+        /// </summary>
+        /// <param name="t">The type to convert.</param>
+        /// <exception cref="NotSupportedException"><paramref name="t"/> is unsupported.</exception>
         public static DataType TypeToDataType(Type t)
```
teo-tsirpanis, line 1107: I don't like this, it restricts us to a one-to-one mapping between .NET types and TileDB data types (it will be a problem once we allow mapping between TileDB and .NET booleans). We might need a `bool IsCompatible<T>(DataType dataType)` method.

sources/TileDB.CSharp/FragmentInfo.cs
```diff
@@ -0,0 +1,681 @@
...
+        private (string Start, string End) GetStringMinimumBoundedRectangle(uint fragmentIndex, uint minimumBoundedRectangleIndex, uint dimensionIndex, DataType dataType)
+        {
+            using var ctxHandle = _ctx.Handle.Acquire();
+            using var handle = _handle.Acquire();
+            ulong startSize64, endSize64;
+            _ctx.handle_error(Methods.tiledb_fragment_info_get_mbr_var_size_from_index(ctxHandle,
+                handle, fragmentIndex, minimumBoundedRectangleIndex, dimensionIndex, &startSize64, &endSize64));
+            int startSize = checked((int)startSize64);
+            int endSize = checked((int)endSize64);
+            using var startBuf = new ScratchBuffer<byte>(startSize, stackalloc byte[128]);
```
ihnorton, line 490: Does this implicitly mean that the maximum size is 128? Or will ScratchBuffer spill to the heap if larger?

sources/TileDB.CSharp/FragmentInfo.cs
```diff
@@ -0,0 +1,681 @@
...
+            byte* mbr = stackalloc byte[sizeof(ulong) * 2];
+            _ctx.handle_error(Methods.tiledb_fragment_info_get_mbr_from_name(ctxHandle,
+                handle, fragmentIndex, minimumBoundedRectangleIndex, ms_dimensionName, mbr));
+
+            var start = Unsafe.ReadUnaligned<T>(mbr);
+            var end = Unsafe.ReadUnaligned<T>(mbr + Unsafe.SizeOf<T>());
+            return (start, end);
+        }
+
+        private (string Start, string End) GetStringMinimumBoundedRectangle(uint fragmentIndex, uint minimumBoundedRectangleIndex, uint dimensionIndex, DataType dataType)
```
ihnorton, line 481: These should probably be restricted to TILEDB_STRING_ASCII for now.

sources/TileDB.CSharp/FragmentInfo.cs
ihnorton: Ok, I see how it works - nice 💯

ihnorton (approved):
One minor nit, otherwise LGTM - please feel free to merge when that one is addressed, or skip if not applicable.

sources/TileDB.CSharp/FragmentInfo.cs
teo-tsirpanis: Why?

sources/TileDB.CSharp/FragmentInfo.cs
teo-tsirpanis: Removed it after seeing the deprecation policy.

sources/TileDB.CSharp/FragmentInfo.cs
ihnorton: That's the only supported var-size type for dimensions. We _should_ throw an error from the C API, but you could give a better/earlier error here. Doesn't need to block merging though.

teo-tsirpanis: If there is no chance for the C API to give us a Unicode string-typed dimension, the relevant decoding code will just not get executed. I will keep it since it might be useful for other string decoding cases.

PR #131, opened 2022-11-01
Add C#-friendly enum value names.

teo-tsirpanis:
The value names of the enums exposed to C# users have the `TILEDB_` prefix and are in all-caps. This PR adds new values with names more friendly to C# and obsoletes the old ones. For example instead of `array.Open(QueryType.TILEDB_WRITE);` we can now write the much better `array.Open(QueryType.Write);`. There will be some warnings guiding users to use the enum valuess, but no breaking changes were introduced.

- [x] TODO: Document the enums.


teo-tsirpanis:
When reviewing this, please make sure that the documentation is correct.

ihnorton (approved):

sources/TileDB.CSharp/Enums.cs
```diff
@@ -1,179 +1,756 @@
...
+        /// </summary>
+        ColumnMajor = tiledb_layout_t.TILEDB_COL_MAJOR,
+        /// <summary>
+        /// Data are in global order.
+        /// </summary>
+        GlobalOrder = tiledb_layout_t.TILEDB_GLOBAL_ORDER,
+        /// <summary>
+        /// Data are unordered.
+        /// </summary>
+        // TODO: Make them more explanatory.
```
shaunrd0, line 531: Just wanted to point out the `TODO` here in case you left it unintentionally. I use them too and accidentally commit them sometimes.

shaunrd0 (approved):
Thanks! Two minor comments around docs but LGTM

sources/TileDB.CSharp/Enums.cs
```diff
@@ -1,179 +1,756 @@
...
+        /// <summary>
+        /// A signed 16-bit integer.
+        /// </summary>
+        Int16 = tiledb_datatype_t.TILEDB_INT16,
+        /// <summary>
+        /// An unsigned 16-bit integer.
+        /// </summary>
+        UInt16 = tiledb_datatype_t.TILEDB_UINT16,
+        /// <summary>
+        /// An unsigned 64-bit integer.
```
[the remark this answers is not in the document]
teo-tsirpanis: Fixed.

sources/TileDB.CSharp/Enums.cs
teo-tsirpanis: I left it intentionally; I still can't figure them out after being explained what they are. I opened [SC-23104](https://app.shortcut.com/tiledb-inc/story/23104/better-explain-in-the-documentation-what-unordered-and-global-order-layouts-are) to update the documentation.

teo-tsirpanis:
Let's wait before merging it. I have to figure out what to do with the members of the `Consts` class and the `EnumUtil` methods that convert between data types and .NET `Type`s; perhaps they deserve a better place.

teo-tsirpanis:
* I deprecated `Constants.TILEDB_VAR_NUM` and replaced it with the more appropriately named and placed `Attribute.VariableSized` and `Dimension.VariableSized`.
* I also deprecated the methods in `EnumUtil` that work with `tiledb_datatype_t`, directing users to use `DataType`.

Let me know if you have any questions.

PR #154, opened 2022-11-27
Don't run nightly builds on weekends.

teo-tsirpanis:


shaunrd0 (approved):
Approved, but I would consider leaving them run for weekends if this is just to align with the work week. In the past I've caught a nightly on a weekend and poked at a branch or read through some logs, but it really just depends what I'm up to.

You can configure notifications for certain GH organizations to send to your work email, if that helps at all. I know for me personally this really helped to organize email alerts.

teo-tsirpanis:
Let's not do it.

PR #157, opened 2022-11-29
Strongly-typed `Config`.

teo-tsirpanis:
[SC-22545](https://app.shortcut.com/tiledb-inc/story/22545/strongly-typed-config-apis)


teo-tsirpanis:
Let's not do this right now.

PR #160, opened 2022-12-01
More quality fixes.

teo-tsirpanis:
This PR:

*
  Adds custom diagnostic IDs[^better-obsoletion] to the recent obsoletions we made, allowing user code to suppress them more granularly. This is the experience users will get when they upgrade:
  image
  If they get too many warnings and don't have the time to fix them immediately they can suppress `TILEDB0001` which will suppress only the warnings about the `TILEDB_` enum values. `TILEDB0001` is also a link, which will lead to more information about the obsoletion.
* Obsoletes all public types in the `TileDB.Interop` namespace (our projects suppress the warnings), paving the way to make them internal in a future release.
* Renames `FileSystemType.Gcs` and `MemFs` to `GoogleCloudStorage` and `InMemory`. These names are much clearer and we can change them as they have not yet shipped.
* Fixes a badly formed piece of documentation in `FragmentInfo.GetFragmentToVacuumUri`.
* ~~Mitigates~~ Reverts a breaking change in the `File.BufferImport` and `BufferExport` methods; these methods were accepting 64-bit unsigned integers until they were changed to native unsigned integers in #105; ~~I added a `ulong` overload that performs a checked conversion to `nuint`.~~

[^better-obsoletion]: https://github.com/dotnet/designs/blob/main/accepted/2020/better-obsoletion/better-obsoletion.md


sources/TileDB.CSharp/Interop/tiledb_walk_order_t.cs
```diff
@@ -1,7 +1,11 @@
 // <auto-generated/>
 
+using System;
+using TileDB.CSharp;
+
 namespace TileDB.Interop
 {
+    [Obsolete(Obsoletions.TileDBInteropMessage + " Use the TileDB.CSharp.WalkOrder type instead.", DiagnosticId = Obsoletions.TileDBInteropDiagId, UrlFormat = Obsoletions.SharedUrlFormat)]
```
shaunrd0, line 8: Still working on relocating some of these to `InteropAugments`? I regenerated using this branch and these were removed by `ClangSharpPInvokeGenerator`.

shaunrd0 (approved):

PR #168, opened 2022-12-05
Add a Release workflow.

teo-tsirpanis:
This PR adds a workflow that runs when a tag gets pushed, packs `TileDB.CSharp` and pushes it to nuget.org. With that the release process gets fully automated, and produces [deterministic NuGet packages compatible with SourceLink](https://mitchelsellers.com/blog/article/net-5-deterministic-builds-source-linking).

Before merging it you have to create a [NuGet API key](https://www.nuget.org/account/apikeys) that can push new versions of `TileDB.CSharp` and add it as a [secret](https://github.com/TileDB-Inc/TileDB-CSharp/settings/secrets/actions) with the name `TILEDB_CSHARP_NUGET_KEY`.

image

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,26 @@
+name: Release
+
+on:
+  release:
+    types:
+      - published
+
+jobs:
+  Release:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Run tests
+        uses: ./tiledb-csharp.yml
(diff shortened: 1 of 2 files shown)


shaunrd0 (approved):
LGTM, approving in advance of adding secret for NuGet key as I lack permissions to do so.

ihnorton:
Key is created

PR #137, opened 2022-11-07
Stop testing 2.11 on nightly builds.

teo-tsirpanis:
Nightly builds on 2.11 are failing because the `tiledb_status_code` function does not exist. This PR modifies `nightly.yml` to stop testing on 2.11.

Fixes #134 #135 #136 #140 #142. See [SC-23174](https://app.shortcut.com/tiledb-inc/story/23174/reconsider-testing-with-older-core-versions).

PR #169, opened 2022-12-05
Deletes unit tests

shaunrd0:
Support for using deletes was added in #130 with regeneration of bindings against 2.12 - this PR adds unit tests for deletes, along with some helper functions in `TestUtil` to reuse common logic for reading / writing to test arrays.


teo-tsirpanis:
Thanks, I will take a beter look at it tomorrow. Until then can you replace all uses of `dynamic` with `object` unless `dynamic` is actually necessary?

tests/TileDB.CSharp.Test/DeletesTest.cs
```diff
@@ -0,0 +1,928 @@
...
+        [DataRow(LayoutType.RowMajor, true, true)]
+        [DataRow(LayoutType.RowMajor, true, false)]
+        [DataRow(LayoutType.RowMajor, false, true)]
+        [DataRow(LayoutType.RowMajor, false, false)]
+        [DataRow(LayoutType.GlobalOrder, true, true)]
+        [DataRow(LayoutType.GlobalOrder, true, false)]
+        [DataRow(LayoutType.GlobalOrder, false, true)]
+        [DataRow(LayoutType.GlobalOrder, false, false)]
+        [DataTestMethod]
+        public void MultipleCells(LayoutType layout, bool legacy, bool duplicates)
```
teo-tsirpanis, line 101: Can you replace the `bool legacy` with `string queryReader`, taking `"legacy"` or `"refactored"`? After #157 it will be easier to switch to an enum.

tests/TileDB.CSharp.Test/DeletesTest.cs
```diff
@@ -0,0 +1,928 @@
...
+
+namespace TileDB.CSharp.Test
+{
+    [TestClass]
+    public class DeletesTest
+    {
+        private static void InitTest(Context ctx, string arrayName, LayoutType layout, bool duplicates)
+        {
+            Console.WriteLine(Assembly.GetAssembly(typeof(TileDB.CSharp.Array))!.FullName);
+            Console.WriteLine($"TileDB Core version: {TileDB.CSharp.CoreUtil.GetCoreLibVersion()}");
```
teo-tsirpanis, line 17: Are these logs always needed to be printed in every test?

Also `Console.WriteLine` is not the right way to log in tests; MSTest seems to have the [`Logger`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.testtools.unittesting.logging.logger) class.

tests/TileDB.CSharp.Test/DeletesTest.cs
```diff
@@ -0,0 +1,928 @@
...
+            var colsCondition = new QueryCondition(ctx);
+            colsCondition.Init("cols", 1, QueryConditionOperatorType.Equal);
+            var rowsCondition = new QueryCondition(ctx);
+            rowsCondition.Init("rows", 1, QueryConditionOperatorType.Equal);
+            var andCondition = colsCondition.Combine(rowsCondition, QueryConditionCombinationOperatorType.And);
+            deleteQuery.SetCondition(andCondition);
+            deleteQuery.Submit();
+            array.Close();
+            Assert.AreEqual(deleteQuery.Status(), QueryStatus.Completed);
+            Console.WriteLine($"Delete single cells status: {deleteQuery.Status()}");
```
teo-tsirpanis, line 147: Logs like this are not needed. You asserted just before that `deleteQuery` is completed; if it's not the test will fail and it will tell you the query's status.

tests/TileDB.CSharp.Test/DeletesTest.cs
```diff
@@ -0,0 +1,928 @@
...
+            string arrayName = "deletes-single-cell";
+            var ctx = Context.GetDefault();
+            ctx.Config().Set("sm.query.sparse_global_order.reader", legacy ? "legacy" : "refactored");
+            ctx.Config().Set("sm.query.sparse_unordered_with_dups.reader", legacy ? "legacy" : "refactored");
+            InitTest(ctx, arrayName, layout, duplicates);
+            var array = new Array(ctx, arrayName);
+
+            // Delete one cell
+            array.Open(QueryType.Delete);
+            var deleteQuery = new Query(ctx, array, QueryType.Delete);
```
teo-tsirpanis, line 137: Can you add `using var` to all TileDB objects? This will dispose them at the end of the method without waiting for the garbage collector.

tests/TileDB.CSharp.Test/DeletesTest.cs
```diff
@@ -0,0 +1,928 @@
...
+    public class DeletesTest
+    {
+        private static void InitTest(Context ctx, string arrayName, LayoutType layout, bool duplicates)
+        {
+            Console.WriteLine(Assembly.GetAssembly(typeof(TileDB.CSharp.Array))!.FullName);
+            Console.WriteLine($"TileDB Core version: {TileDB.CSharp.CoreUtil.GetCoreLibVersion()}");
+            if (!TestUtil.VersionMinimum(2, 12))
+            {
+                return;
+            }
```
teo-tsirpanis, line 21:
```suggestion
```

You can delete this; we no longer test with 2.11 after #137.

tests/TileDB.CSharp.Test/DeletesTest.cs
```diff
@@ -0,0 +1,928 @@
...
+
+            var elementCount = (int)bufferElements["a1"].Item1;
+            var offsetCount = (int)bufferElements["a1"].Item2!;
+            var a1Data = new List<string>();
+            for (int i = 0; i < offsetCount; i++)
+            {
+                int cellSize = i == offsetCount - 1
+                    ? elementCount - (int)readOffsetBuffers["a1"][i]
+                    : (int)readOffsetBuffers["a1"][i + 1] - (int)readOffsetBuffers["a1"][i];
+                a1Data.Add(new (Encoding.ASCII.GetChars(readBuffers["a1"], (int)readOffsetBuffers["a1"][i],
```
teo-tsirpanis, line 889: After removing `dynamic` you will have to cast `readBuffers["a1"]` to `byte[]`.

tests/TileDB.CSharp.Test/TestUtil.cs
```diff
@@ -84,6 +87,148 @@ public static void CreateArray(Context ctx, string name, ArraySchema schema)
...
+                Console.WriteLine($"Expected {buffer.Key}: {string.Join(", ", buffer.Value)}" +
+                                  $"\n  Actual {buffer.Key}: {string.Join(", ", actual[buffer.Key])}");
+                if (duplicates)
+                {
+                    CollectionAssert.AreEquivalent(buffer.Value, actual[buffer.Key]);
+                }
+                else
+                {
+                    CollectionAssert.AreEqual(buffer.Value, actual[buffer.Key]);
+                }
```
teo-tsirpanis, line 227: Hmm, you are actually making use of `dynamic` here. 🤔

teo-tsirpanis (changes requested):

teo-tsirpanis:
Can you add tests for `QueryConditionCombinationOperatorType.Not`?

tests/TileDB.CSharp.Test/DeletesTest.cs
```diff
@@ -0,0 +1,896 @@
...
+            using var deleteQuery = new Query(ctx, array, QueryType.Delete);
+            using var colsCondition = new QueryCondition(ctx);
+            colsCondition.Init("cols", 1, QueryConditionOperatorType.Equal);
+            using var rowsCondition = new QueryCondition(ctx);
+            rowsCondition.Init("rows", 1, QueryConditionOperatorType.Equal);
+            using var andCondition = colsCondition.Combine(rowsCondition, QueryConditionCombinationOperatorType.And);
+            deleteQuery.SetCondition(andCondition);
+            deleteQuery.Submit();
+            array.Close();
+            Assert.AreEqual(deleteQuery.Status(), QueryStatus.Completed);
```
teo-tsirpanis, line 195: The order of parameters must be swapped; first goes the expected value and then the actual. It should be caught by SonarSource; opened SonarSource/sonar-dotnet#6547.

shaunrd0: I swapped the order of these assertions but it looks like the NOT operator isn't supported yet for [QueryConditionCombinationOp](https://github.com/TileDB-Inc/TileDB/blob/dev/tiledb/sm/query/query_condition.cc#L2122)

teo-tsirpanis (changes requested):
Can you ensure that the TileDB objects, particularly the `Query`ies are disposed? I added a `keepOpen` parameter to `ReadArray` and `WriteArray`, you should add a `keepOpen: false` to calls of these methods where the query object is not used, and add `using` to those where it gets used.

tests/TileDB.CSharp.Test/DeletesTest.cs
```diff
@@ -0,0 +1,894 @@
...
+            InitTest(ctx, arrayName, layout, duplicates);
+            using var array = new Array(ctx, arrayName);
+
+            // Delete one cell
+            array.Open(QueryType.Delete);
+            using var deleteQuery = new Query(ctx, array, QueryType.Delete);
+            using var colsCondition = new QueryCondition(ctx);
+            colsCondition.Init("cols", 1, QueryConditionOperatorType.Equal);
+            using var rowsCondition = new QueryCondition(ctx);
+            rowsCondition.Init("rows", 1, QueryConditionOperatorType.Equal);
```
teo-tsirpanis, line 135: Can you use `QueryCondition.Create`?

teo-tsirpanis (approved):

PR #190, opened 2022-12-21
Invalidate safe handles that failed to be created.

teo-tsirpanis:
This will allow the GC to skip finalization and collect them sooner.

### DIFF
--- a/sources/TileDB.CSharp/ArraySchema.cs
+++ b/sources/TileDB.CSharp/ArraySchema.cs
@@ -239,6 +239,10 @@ namespace TileDB.CSharp
                 {
                     handle.InitHandle(filter_list_p);
                 }
+                else
+                {
+                    handle.SetHandleAsInvalid();
+                }
             }
 
             return new FilterList(_ctx, handle);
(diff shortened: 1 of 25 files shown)


sources/TileDB.CSharp/ArraySchema.cs
```diff
@@ -239,6 +239,10 @@ public FilterList CoordsFilterList()
                 {
                     handle.InitHandle(filter_list_p);
                 }
+                else
+                {
+                    handle.SetHandleAsInvalid();
+                }
```
teo-tsirpanis, line 245: To make it easier to understand, `successful` being false means that an exception was thrown during the handle's creation. We would not be able to return the handle, but it would still exist.

Before this change the handle would stay alive for two garbage collections so that `ReleaseHandle` can be called in the finalizer. But the handle's value is `null` and we don't have anything to relese. `SetHandleAsInvalid` skips this finalization, allowing the handle to be collected after just one garbage collection, and also prevents it from being used (in case we accidentally return such handle).

ihnorton (approved):

PR #191, opened 2022-12-21
Add `TileDBException` and obsolete `Context.ErrorHappened` and `ErrorException`.

teo-tsirpanis:
Until now, when a native API call fails, we throw a plain `Exception`, giving little ways for consumers to distinguish between TileDB errors and anything else.

This PR introduces a new exception type called `TileDBException` that gets thrown when a native API call fails, and corresponds to the `TileDBError` type that C++, Python and Java have. Since we throw a more derived type of exception, [this is not a breaking change](https://learn.microsoft.com/en-us/dotnet/core/compatibility/#exceptions).

I also made the `Context.ErrorHappened` ~~event useful; as explained in [SC-23974](https://app.shortcut.com/tiledb-inc/story/23974/context-errorhappened-is-useless), the default value we give to it throws and prevents user-supplied events to be invoked. We might want to just~~ obsolete and ignore it; I still prefer it.

### DIFF
--- a/sources/TileDB.CSharp/ErrorHandling.cs
+++ b/sources/TileDB.CSharp/ErrorHandling.cs
@@ -1,5 +1,4 @@
-﻿using System;
-using System.Diagnostics;
+﻿using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using TileDB.Interop;
 
@@ -20,7 +19,7 @@ namespace TileDB.CSharp
         {
             if (status == (int)Status.TILEDB_OK)
             {
-                Methods.tiledb_error_free(error);
+                Debug.Assert(*error is null);
                 return;
             }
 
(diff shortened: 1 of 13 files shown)


sources/TileDB.CSharp/ErrorHandling.cs
```diff
@@ -10,7 +12,29 @@ public static void ThrowOnError(int errorCode)
...
             }
-            throw new Exception($"Operation failed with error code {errorCode}.");
+            throw new TileDBException($"Operation failed with error code {errorCode}.") { StatusCode = errorCode };
+        }
+
+        public static unsafe void CheckLastError(tiledb_error_t** error, int status)
+        {
+            if (status == (int)Status.TILEDB_OK)
+            {
+                Debug.Assert(*error is null);
```
teo-tsirpanis, line 22: Is that correct? Is there a case when the status is `OK` and `*error != null`?

ihnorton (approved):

PR #193, opened 2022-12-22
Do not use `GCHandle`s for brief pinning of objects.

teo-tsirpanis:
This PR removes most uses of `GCHandle`s and replaces them with directly using pointers when passing values on the stack to native code, or with the `fixed` statement when pinning values on the heap.

The remaining uses of `GCHandle`s are in the bindings example (that will eventually be removed), the deprecated `TileDBBuffer` class, the soon-to-be-deprecated subarray methods on `Query`, and the long-term pinning of query buffers and keeping alive the `Query` for the duration of an async submit (the only valid use casees).

I also took the opportunity to clean-up the files I touched and deprecate the `Dimension.Create` overload that takes an array; only the first two values are actually used; the overload that explicitly takes the lower and upper bounds better represents the method's intent.

[SC-23109](https://app.shortcut.com/tiledb-inc/story/23109/do-not-use-gchandles-for-brief-pinning-of-objects)

### DIFF
--- a/sources/TileDB.CSharp/File.cs
+++ b/sources/TileDB.CSharp/File.cs
@@ -1,3 +1,4 @@
+using System.ComponentModel;
 using TileDB.CSharp.Marshalling.SafeHandles;
 using TileDB.Interop;
 
@@ -100,6 +101,7 @@ namespace TileDB.CSharp
             return size;
         }
 
+        [EditorBrowsable(EditorBrowsableState.Never)]
         public string MIMETypeToStr(MIMEType mimeType)
         {
             var tiledb_mime_type = (tiledb_mime_type_t)mimeType;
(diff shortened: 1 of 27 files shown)


sources/TileDB.CSharp/File.cs
```diff
@@ -61,35 +56,31 @@ public void URIExport(string fileURI, string arrayURI)
 
         public void BufferImport<T>(string arrayURI, T value, ulong size, MIMEType mimeType) where T : struct
         {
+            ErrorHandling.ThrowIfManagedType<T>();
             using var ctxHandle = _ctx.Handle.Acquire();
             using var ms_arrayURI = new MarshaledString(arrayURI);
-            var data = new[] { value };
-            var dataGcHandle = GCHandle.Alloc(data, GCHandleType.Pinned);
```
teo-tsirpanis, line 67: We did not free this `GCHandle` and had a memory leak.

PR #205, opened 2022-12-29
Optimize and clean-up the `Query` class, and support setting buffers from more sources.

teo-tsirpanis:
Apart from trivial clean-ups and fixes, the most important change in this PR is a refactoring of how the query buffers are managed.

Instead of `GCHandle` we use the newer `MemoryHandle` type which is more flexible and supports arbitrary pointers that might or might not be pinned. On top of that I added overloads to `Set***Buffer` that accept buffers from `Memory<T>` (a safe type that can repserent buffers like an array or part of an array). These overloads are essential for enabling user code to optimize performance and memory efficiency. I also changed the order of setting a buffer and unpinning the previous one to be more resilient to errors the Core might report.

[SC-24176](https://app.shortcut.com/tiledb-inc/story/24176/allow-query-buffers-from-more-sources)

### DIFF
--- a/sources/TileDB.CSharp/Query.cs
+++ b/sources/TileDB.CSharp/Query.cs
@@ -278,13 +278,13 @@ namespace TileDB.CSharp
             SetDataBuffer(name, PinArray(data), (ulong)data.Length * (ulong)sizeof(T));
         }
 
-        private void SetDataBuffer(string name, MemoryHandle memoryHandle, ulong size)
+        private void SetDataBuffer(string name, MemoryHandle memoryHandle, ulong byteSize)
         {
             BufferHandle? handle = null;
             bool successful = false;
             try
             {
-                handle = new BufferHandle(ref memoryHandle, size);
+                handle = new BufferHandle(ref memoryHandle, byteSize);
 
                 SetDataBuffer(name, handle.DataPointer, handle.SizePointer);
 
@@ -321,7 +321,7 @@ namespace TileDB.CSharp
                 throw new ArgumentException("Query.set_offsets_buffer, buffer is null or empty!");
             }
 
-            SetOffsetsBuffer(name, PinArray(data), (ulong)data.Length * sizeof(ulong));
+            SetOffsetsBuffer(name, PinArray(data), (ulong)data.Length);
         }
 
         private void SetOffsetsBuffer(string name, MemoryHandle memoryHandle, ulong size)
@@ -330,7 +330,7 @@ namespace TileDB.CSharp
             bool successful = false;
             try
             {
-                handle = new BufferHandle(ref memoryHandle, size);
+                handle = new BufferHandle(ref memoryHandle, size * sizeof(ulong));
 
                 SetOffsetsBuffer(name, (ulong*)handle.DataPointer, handle.SizePointer);
 
--- a/sources/TileDB.CSharp/Query.cs
+++ b/sources/TileDB.CSharp/Query.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.ComponentModel;
+using System.Diagnostics;
 using System.Linq;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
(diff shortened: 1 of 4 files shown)


sources/TileDB.CSharp/Query.cs
```diff
@@ -246,83 +220,430 @@ public Config Config()
...
+                handle = new BufferHandle(ref memoryHandle, byteSize);
+
+                SetDataBufferCore(name, handle.DataPointer, handle.SizePointer);
+
+                AddOrReplaceBufferHandle(_dataBufferHandles, name, handle);
+                successful = true;
+            }
+            finally
             {
-                SetDataBuffer<byte>(name, System.Array.ConvertAll(boolData as bool[], d => d ? (byte)1 : (byte)0));
```
teo-tsirpanis, line 267: We might need to discuss this. The Core does not set any expectations about the definition of bool, particularly around the values that are not one or zero.

I removed it because there are correctness concerns besides performance ones. Imagine we set a data buffer to a `bool[]`, we submit the query and change our `bool[]` to contain the new data. _The Core will not see the new data._

And also since in C# we get bools that are not one or zero only by using `unsafe` code I don't think this is very impactful.

sources/TileDB.CSharp/Query.cs
```diff
@@ -1058,66 +1350,79 @@ private DataType GetDataType(string name)
...
-        /// <summary>
-        /// Add offset buffer handle.
-        /// </summary>
-        /// <param name="name"></param>
-        /// <param name="handle"></param>
-        /// <param name="size"></param>
-        private void AddOffsetsBufferHandle(string name, GCHandle handle, ulong size)
-        {
-            if (_offsetsBufferHandles.ContainsKey(name))
+            public BufferHandle(ref MemoryHandle handle, ulong size)
```
teo-tsirpanis, line 1397: We pass the `MemoryHandle` by reference to ensure it exists in only one place in memory and avoid disposing it twice. Disposing `default(MemoryHandle)` does nothing.

sources/TileDB.CSharp/Query.cs
```diff
@@ -987,34 +1279,32 @@ private void CheckDataType<T>(DataType dataType)
...
+                using var dimension = domain.Dimension(name);
+                return dimension.Type();
             }
-            else if (name == "__coords")
-            {
-                return _array.Schema().Domain().Type();
-            }
-            else
+
+            if (name == "__coords")
```
teo-tsirpanis, line 1300: Isn't is about that deprecated zipped coords feature? Do we still need it?

teo-tsirpanis (commented):
I added comments about some areas of interest.

teo-tsirpanis:
And with that, my Pull Request flurry is over. 😅 Apart from small or urgent ones, future PRs will be opened after all existing non-draft ones get addressed.

teo-tsirpanis:
This is an overview of the proposed API shape:

```csharp
public sealed class Query
{
    // Existing APIs
    public void SetDataBuffer<T>(string name, T[] data);
    public void SetOffsetsBuffer(string name, ulong[] data);
    public void SetValidityBuffer(string name, byte[] data);

    // New APIs

    // These APIs take Memory<T> instead of T[].
    // If we don't care about 64-bit lengths, we can expose only these and call it a day.
    public void SetDataBuffer<T>(string name, Memory<T> data);
    public void SetOffsetsBuffer(string name, Memory<ulong> data);
    public void SetValidityBuffer(string name, Memory<byte> data);

    // These APIs take a pointer and a length. They support 64-bit lengths.
    // Discounting the 64-bit length, they could be implemented on top of
    // the Memory<T>-accepting overloads, but it will require an object allocation.
    public unsafe void SetDataBuffer<T>(string name, T* data, ulong size);
    public unsafe void SetOffsetsBuffer(string name, ulong* data, ulong size);
    public unsafe void SetValidityBuffer(string name, byte* data, ulong size);

    // These APIs take a ReadOnlyMemory<T> and work only on write queries.
    // We could not add them, there are few cases where we have only a ReadOnlyMemory,
    // and users can use the unsafe MemoryMarshal.AsMemory method to convert it to a Memory.
    // And as for the name, SetDataReadOnlyBuffer seems better. 🤔
    public void SetDataWriteBuffer<T>(string name, ReadOnlyMemory<T> data);
    public void SetOffsetsWriteBuffer(string name, ReadOnlyMemory<ulong> data);
    public void SetValidityWriteBuffer(string name, ReadOnlyMemory<byte> data);

    // These APIs take a pointer and a length in bytes.
    // They are unsafe because they do not validate the type.
    // They might have some use case in dynamic scenarios where the type is
    // not known at compile time, but there are not many possible types, user code
    // could switch over the possible data types (u?int(8|16|32|64), bool, float and double).
    // The overload accepting a MemoryHandle allows the Query class to manage the lifetime and unpin.
    public unsafe void UnsafeSetDataBuffer(string name, void* data, ulong byteSize);
    public void UnsafeSetDataBuffer(string name, MemoryHandle data, ulong byteSize);

    // Withdrawn APIs

    // I used to publicly expose overloads for the offsets and validity buffers taking
    // MemoryHandle, but they are not really useful. If you have a managed buffer just
    // use Memory; managed buffers cannot be more than 2GB long, and if you have an
    // unmanaged buffer use pointers.
    public void UnsafeSetOffsetsBuffer(string name, MemoryHandle data, ulong size);
    public void UnsafeSetValidityBuffer(string name, MemoryHandle data, ulong size);
}
```

sources/TileDB.CSharp/Query.cs
```diff
@@ -208,26 +182,47 @@ public void SetConfig(Config config)
...
+                if (successful)
+                {
+                    handle.InitHandle(config);
+                }
+                else
+                {
+                    handle.SetHandleAsInvalid();
+                }
+            }
+            return new Config(handle);
```
teo-tsirpanis, line 206: Is there a reason we weren't doing that before?

Shelnutt2:
@teo-tsirpanis lets mark as deprecated but not remove the APIs. We typically use a deprecation cycle of at minimum two releases (minor) releases to ensure users have time to migrate off.

teo-tsirpanis:
@Shelnutt2 we won't remove any shipped API. The two withdrawn APIs at the end are new APIs that originally were public but I changed my mind and made them internal.

Shelnutt2:
> @Shelnutt2 we won't remove any shipped API. The two withdrawn APIs at the end are new APIs that originally were public but I changed my mind and made them internal.

Since they were public and now they are not, this means if someone had used them their code will break on upgrade of the library, is that correct? If so that is the same issue as completely removing the API and we need to go through a deprecation cycle.

teo-tsirpanis:
I wasn't clear enough. They were public _initially in this PR_ and now are not. That's what I meant with the "we won't remove any shipped API".

teo-tsirpanis:
I renamed `Set***WriteBuffer` to `Set***ReadOnlyBuffer`.

ihnorton (approved):

PR #217, opened 2023-01-09
Obsolete `ConsolidateMetadata`.

teo-tsirpanis:
It calls the deprecated `tiledb_array_consolidate_metadata` Core method.


sources/TileDB.CSharp/Obsoletions.cs
```diff
@@ -29,5 +29,8 @@ internal static class Obsoletions
 
         public const string QueryConditionCombineMessage = "The QueryCondition.Combine method is obsolete. Use the '&', '|' and '!' operators instead.";
         public const string QueryConditionCombineDiagId = "TILEDB0008";
+
+        public const string ConsolidateMetadataMessage = "The Array.ConsolidateMetadata and ArrayMetadata.ConsolidateMetadata methods are obsolete. Do a regular consolidation with the config value 'sm.consolidation.mode' set to 'array_meta' instead.";
```
shaunrd0, line 33: Just a nit, but I think `Prefer Array.Consolidate with the config value 'sm.consolidation.mode' set to 'array_meta' instead.` or something similar would give a more direct resolution to users.

shaunrd0 (approved):
Just a small nit on error message wording, but LGTM :+1:

sources/TileDB.CSharp/Obsoletions.cs
teo-tsirpanis: Good idea, I replaced "prefer" with "call".

PR #234, opened 2023-01-23
Support query conditions that compare an attribute to null.

teo-tsirpanis:
[SC-24795](https://app.shortcut.com/tiledb-inc/story/24795/support-comparing-with-null-in-query-conditions)

This PR adds new APIs to `QueryCondition`:

```csharp
public static QueryCondition CreateIsNull(Context ctx, string attribute_name);
public static QueryCondition CreateIsNotNull(Context ctx, string attribute_name);
```

that allow creating query conditions that compare whether a nullable attribute is null or not.

As for the naming, another idea is `CreateEqualsNull` or `CreateNotEqualsNull` but it's longer and `x is null` is valid C# since quite some time.


ihnorton (approved):
LGTM, can you please add a small test before merging though. Thanks!

teo-tsirpanis:
I added tests. Let's see what CI thinks.

PR #240, opened 2023-01-30
Remove obsolete APIs.

teo-tsirpanis:
Mimicking the Core's deprecation policy, this PR removes all C# APIs marked with `[Obsolete]` (they were obsoleted on or before 5.3.0, 5.4.0 did not obsolete anything and 5.5.0 is coming next, coinciding with Core 2.15.0 which will also start removing). The obsoletions were split to three commits, in descending order of necessity to remove and ascending order of impact.

1. The first commit removes `Query.SubmitAsync` and its related APIs. When we want to bring it back, it will have an improved shape and we must remove the old one first.
2. The second commit removes almost all other APIs. All had documented straightforward replacements.
3. The third commit removes the `Char`, `StringUcs2` and `StringUcs4` data types (any did not receive a C# name and was removed by item 2). The removal of these data types might affect some cases so I placed it separately.

The low-level interop API was not removed; there is a bug in ClangSharpPInvokeGenerator that got fixed in future versions, which we cannot upgrade to because of dotnet/ClangSharp#413.

I can also see the argument that we (the C# API) did not specifically say that these APIs will be one day removed, and we should announce it and wait two releases from now, but obsoletions in the C# API have more user-visible migration paths than the C API; I cannot imagine anyone in a new version still using the old APIs.


ihnorton (approved):
LGTM, but please wait until tomorrow to merge in case @Shelnutt2 has any feedback.

teo-tsirpanis:
24 hours have passed without additional feedback, merging…

PR #251, opened 2023-02-21
Use SemVer 2.0 versions for pre-release native packages.

teo-tsirpanis:
Pre-release Core versions have the form `<major>.<minor>.0-rc<rc_number>`. When publishing them to NuGet, adding a dot between the `-rc` and the RC number would make the package version [compatible with Semantic Versioning 2.0](https://learn.microsoft.com/en-us/nuget/concepts/package-versioning#semantic-versioning-200), enabling things like `rc.10` being greater than `rc.1`. This PR adds this dot.


teo-tsirpanis:
Let's not do this; we haven't published an RC on NuGet before.

PR #254, opened 2023-03-03
Update Core to 2.15.0 and improve the bindings generation process.

teo-tsirpanis:
This PR updates the bindings generation process, finally ensuring that all APIs are present. Bindings were also regenerated for version 2.15.0

~~Depends on Core version 2.15.0 being released.~~ Also depends on some unreleased ClangSharpPInvokeGenerator patches.


teo-tsirpanis:
Tests fail because `tiledb_fragment_info_get_fragment_name_v2` cannot be found; this is expected.

examples/bindings/QuickstartSparseString/Program.cs
```diff
@@ -1,340 +0,0 @@
-using System;
-using System.IO;
-using System.Runtime.InteropServices;
-using System.Text;
-using TileDB.CSharp;
-using TileDB.CSharp.Examples;
-using TileDB.Interop;
-
-namespace QuickstartSparseString
```
ihnorton, line 9: Is this removed because of the change to `internal` linkage for the generated classes? How could one write a demo/test against the generated C API wrapper code now?

ihnorton (approved):

examples/bindings/QuickstartSparseString/Program.cs
teo-tsirpanis: By directly importing the generated sources in that demo/test project. This can be done by adding something like the following lines in the project file:

```xml
<ItemGroup>
  <Compile Include="sources/TileDB.CSharp/Interop/*.cs" />
</ItemGroup>
```

PR #258, opened 2023-03-09
Add experimental dimension label APIs

teo-tsirpanis:
### TODO
- [x] Add `Subarray` getter APIs.
- [x] Add examples.
- [ ] Add tests.

FYI @jp-dark


sources/TileDB.CSharp/FragmentInfo.cs
```diff
@@ -651,7 +651,7 @@ private static void ValidateDomainType<T>(DataType dataType)
             {
                 ThrowHelpers.ThrowStringTypeMismatch(dataType);
             }
-            if (dataType != EnumUtil.TypeToDataType(typeof(T)))
+            if (EnumUtil.DataTypeToType(dataType) != typeof(T))
```
teo-tsirpanis, line 654: This fixes a bug in our type checking logic that manifested when using date/time types in the dimension labels example.

There is a many-to-one mapping between a TileDB data type and a .NET type. `long` maps by default to `TILEDB_INT64` but there are other possible TileDB data types that map to `long`, such as `TILEDB_DATETIME_*S`. We were previously "widening" the .NET type into its default TileDB data type, which caused mismatches. The correct way is to "narrow" the TileDB data type into the .NET type.

teo-tsirpanis:
[SC-26013](https://app.shortcut.com/tiledb-inc/story/26013/expose-the-dimension-label-api-from-c)

sources/TileDB.CSharp/FragmentInfo.cs
teo-tsirpanis: Fixed in #291.

PR #260, opened 2023-03-15
Bump version to 5.5.0.

teo-tsirpanis:


ihnorton (approved):
Where do we set the pin to 2.15? Does that happen directly in nuget or ???

teo-tsirpanis:
https://github.com/TileDB-Inc/TileDB-CSharp/blob/30f6075eb3143fd7a493978bb43d96280b46b3cd/Directory.Packages.props#L5

teo-tsirpanis:
This file is automatically imported to all projects in the repository by MSBuild.

ihnorton:
Got it, thanks!

PR #261, opened 2023-04-07
Improve string decoding; use UTF-8 by default.

teo-tsirpanis:
This PR does three things:

*
  It changes the default encoding for incoming strings from the Core from ASCII to UTF-8. It is needed for TileDB-Inc/TileDB#3987 and is more flexible in general.

  If `Encoding.UTF8` encounters invalid byte sequences, it converts them to `U+FFFD � REPLACEMENT CHARACTER`. If `Encoding.ASCII` encounters invalid bytes, it converts them to question marks (thought it would throw, turns out it does not).

* It uses the correct encoding in cases we know it (metadata and attribute fill values).

* It changes the encoding for string metadata values to UTF-8. Metadata from existing arrays with different encodings will be correctly handled (from the above bullet point).

### DIFF
--- a/tests/TileDB.CSharp.Test/ArrayMetadataTest.cs
+++ b/tests/TileDB.CSharp.Test/ArrayMetadataTest.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 using System.IO;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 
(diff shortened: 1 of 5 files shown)


tests/TileDB.CSharp.Test/ArrayMetadataTest.cs
```diff
@@ -74,7 +74,7 @@ private void writeArrayMetadata(string tmpArrayPath)
 
             array.PutMetadata<float>("key4", new float[] { 25.1f, 26.2f, 27.3f, 28.4f });
 
-            array.PutMetadata("key5", "This is TileDB array metadata");
+            array.PutMetadata("key5", "This is TileDB array metadata, that supports Unicode characters! 🥳");
```
ihnorton, line 77: 😄

ihnorton (approved):

PR #270, opened 2023-09-06
Enable package validation.

teo-tsirpanis:
This PR uses [the .NET SDK's package validation feature](https://learn.microsoft.com/en-us/dotnet/fundamentals/package-validation/overview) to validate that `TileDB.CSharp` does not have any backwards-incompatible changes, compared to a published version of the package (currently 5.5.0). If we need to introduce a breaking change, we will have to [explicitly suppress the error](https://learn.microsoft.com/en-us/dotnet/fundamentals/package-validation/overview#suppress-compatibility-errors).

I created an additional CI job that runs `dotnet pack` which will trigger the analysis, and will test it by trying to introduce a breaking change in the PR.

### DIFF
--- a/.github/workflows/tiledb-csharp.yml
+++ b/.github/workflows/tiledb-csharp.yml
@@ -9,6 +9,28 @@ on:
   workflow_call:
 
 jobs:
+  Validate-Package:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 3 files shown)


ihnorton (approved):
💯 thanks

PR #278, opened 2023-09-22
Add back removed APIs from the `TileDB.Interop` namespace.

teo-tsirpanis:
They were temporarily reintroduced and immediately deprecated with a new diagnostic ID. They will be removed in the 5.9.x cycle.


ihnorton (approved):
Thanks

PR #284, opened 2023-10-02
Add support for enumerations.

teo-tsirpanis:
[SC-30204](https://app.shortcut.com/tiledb-inc/story/30204/csharp-support-for-enumerated-types)

### TODO

- [x] Add tests.


sources/TileDB.CSharp/Enumeration.cs
```diff
@@ -0,0 +1,330 @@
...
+        /// <summary>
+        /// Disposes the <see cref="Enumeration"/>.
+        /// </summary>
+        public void Dispose()
+        {
+            _handle.Dispose();
+        }
+
+        /// <summary>
+        /// Returns the number of values for each member of the <see cref="Enumeration"/>.
```
ihnorton, line 121: Please clarify this, and also add a note that this is equivalent to cell_val_num

ihnorton (approved):
Minor nit, otherwise LGTM - thanks!

sources/TileDB.CSharp/Enumeration.cs
teo-tsirpanis: Done.

PR #291, opened 2023-10-11
Fix generic type validation to support datetimes and non-ASCII strings.

teo-tsirpanis:
[SC-24612](https://app.shortcut.com/tiledb-inc/story/26412/reconsider-the-rules-of-generic-type-validation)

The C# API's rules around type validation in its generic methods have some defects that prevent datetime types and non-ASCII strings from being used. This PR improves the type validation to fix these cases.

### Summary of changes

* When performing type validation, instead of converting the .NET `Type` to a TileDB `DataType` and compare them, we do the opposite. This happens because (with few exceptions), the mapping between .NET types and TileDB data types is one-to-many, and so by "narrowing down" we can validate that datetime data types match with `long`, while `long` maps to `DataType.Int64`.
  * The validation logic was centralized in the `ErrorHandling.CheckDataType`. It accepts a generic type parameter and a `DataType` regular parameter, and fails if they are incompatible.
* There are two special cases for validation: `DataType.Boolean` and `StringUtf16`. Besides `byte` and `ushort`, they can also match with `bool` and `char` (in .NET it's 2 bytes long) respectively.
  * We could do the same with `StringUtf32` and `System.Text.Rune`, but the Core must validate that UTF-32 strings contain valid Unicode codepoints (if it already does, please let me know).
* _Behavior breaking change:_ The `EnumUtil.DataTypeToType` method mapped all string `DataType`s to `sbyte`, which is an obviously wrong behavior, and inconsistent with what the C++ API does. This PR changes it to map ASCII and UTF-8 strings to `byte`, UTF-16 strings to `ushort`, and UTF-32 strings to `uint`.
  * In case of `DataType`s with multiple valid types (like `StringUtf16` and `Boolean`), `DataTypeToType` returns the matching _numeric_ type.
* _Minor behavior breaking change:_ Some generic methods throw `ArgumentException` which is more appropriate than `InvalidOperationException`. Not commonly expected to hit, but worth mentioning.

### DIFF
--- a/docs/obsoletions.md
+++ b/docs/obsoletions.md
@@ -7,7 +7,7 @@ Following [the deprecation policy of TileDB Embedded][core-deprecation], obsolet
 |Diagnostic codes|Deprecated in version|Removed in version|
 |----------------|---------------------|------------------|
 |[`TILEDB0001`](#TILEDB0001) …[`TILEDB0011`](#TILEDB0011)|5.3.0|5.5.0|
-|[`TILEDB0012`](#TILEDB0012) …[`TILEDB0012`](#TILEDB0012)|5.7.0|5.9.0|
+|[`TILEDB0012`](#TILEDB0012) …[`TILEDB0013`](#TILEDB0013)|5.7.0|5.9.0|
 
 ## `TILEDB0001` - Enum value names that start with `TILEDB_` were replaced with C#-friendly names.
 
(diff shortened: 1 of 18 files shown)


teo-tsirpanis:
The failing examples in #258 succeed after this.

ihnorton:
> Behavior breaking change: The EnumUtil.DataTypeToType method

Is there any way that external users could be relying on this behavior?

teo-tsirpanis:
Besides the validation we do, one other use case I could find for this method is to dynamically construct an array of that type using reflection, as part of processing TileDB arrays of unknown schemata, but there are better ways of doing this.

Another thing we could do is make an internal edition of this method with the updated behavior and deprecate the public `TypeToDataType` and `DataTypeToType` methods, keeping the existing behavior in the meantime.

ihnorton (commented):
> Another thing we could do is make an internal edition of this method with the updated behavior and deprecate the public TypeToDataType and DataTypeToType methods, keeping the existing behavior in the meantime.

+1, let's do it that way, thanks 💪

ihnorton (approved):

ihnorton:
Thanks

PR #299, opened 2023-10-20
Add an example for dataframe-like arrays.

teo-tsirpanis:
[SC-35380](https://app.shortcut.com/tiledb-inc/story/35380/provide-c-example-for-dataframe-like-arrays)


shaunrd0 (approved):
Since this was a user request adding in some short comments wouldn't hurt, but the example is easy enough to follow as-is so LGTM :+1:

PR #358, opened 2023-12-22
Disable AVX2 on nightly builds.

teo-tsirpanis:
Because building the Core and running the C# tests happens on separate jobs, they might run on a different machine. If the building machine supports AVX2 and the testing machine does not, the latter will fail with an invalid opcode error.

This is my latest theory on why the macOS nightly builds often fail, especially since nightly builds on other TileDB projects also fail on macOS. With this PR we always disable AVX2 when building the Core.

If this PR gets merged and we don't get a related nightly failure for two weeks, I will consider the failures fixed and will close all relevant issues.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -20,7 +20,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
         include:
-          - BOOTSTRAP: ../bootstrap --enable-s3 --enable-serialization --force-build-all-deps --enable-vcpkg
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly.yml
```diff
@@ -20,7 +20,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
         include:
-          - BOOTSTRAP: ../bootstrap --enable-s3 --enable-serialization --force-build-all-deps --enable-vcpkg
```
teo-tsirpanis, line 23: `--force-build-all-deps` has no effect with vcpkg enabled.

ihnorton (approved):

PR #362, opened 2024-01-08
Update to the .NET 8 SDK and use some new C# features.

teo-tsirpanis:
> [!IMPORTANT]
> Please review this with whitespace diffs turned off.

Similar to #141, this PR updates the required version of the .NET SDK to .NET 8, and uses some new and less new C# language features along the way. We update the SDK primarily because .NET 7 will go out of support in May 2024, and to enable updating the ClangSharpPInvokeGenerator whose latest version runs on .NET 8.

Like the previous PR, this does not affect the C# API's runtime targeting of .NET 5. .NET 8 will be required only to build it from source.


shaunrd0 (approved):
LGTM :+1:

ihnorton (approved):

PR #379, opened 2024-02-15
Update Core to 2.20.

teo-tsirpanis:


teo-tsirpanis:
`ls_recursive` validated locally by running this:

```csharp
using var config = new Config();
config.Set("vfs.s3.aws_access_key_id", "[REDACTED]");
config.Set("vfs.s3.aws_secret_access_key", "[REDACTED]");
using var ctx = new Context(config);
using var vfs = new VFS(ctx);
vfs.VisitChildrenRecursive("s3://<bucket>/<dir>", (uri, size, _) => {
    Console.WriteLine($"{uri} ({size} bytes)");
    return true;
}, 0);
```

sources/TileDB.CSharp/VFS.cs
```diff
@@ -391,6 +427,38 @@ public void VisitChildren<T>(string uri, Func<string, T, bool> callback, T callb
         callbackData.Exception?.Throw();
     }
 
+    /// <summary>
+    /// Visits all files of a directory and its subdirectories recursively.
```
shaunrd0, line 431: Might be worth mentioning recursive listing is only supported on S3 since this could be the first docs a user reads.

shaunrd0 (approved):

sources/TileDB.CSharp/VFS.cs
teo-tsirpanis: Added a note.

.github/workflows/nightly.yml
```diff
@@ -67,7 +67,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        tag: [release-2.19, dev]
+        tag: [release-2.20, dev]
```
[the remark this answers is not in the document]
teo-tsirpanis: Oops, thanks for spotting!

PR #388, opened 2024-02-26
Fix nightly builds.

teo-tsirpanis:
We build with Ninja, all cores enabled and the superbuild disabled.


teo-tsirpanis:
[Nightlies succeed on my fork.](https://github.com/teo-tsirpanis/TileDB-CSharp/actions/runs/8082177950)

PR #417, opened 2024-04-26
Revert floating the Core's version in tests and fix CI.

teo-tsirpanis:
Fixes nightly builds.

For some backstory, the packaging system of the C# API enables Core patches to be released without any changes to the C# API package. This has the downside that by default the earliest (`.0`) Core patch gets used, which means that to update the Core users have to install the `TileDB.Native` package themselves.

c3f0ace9525cfb401721c032ccaaa6ba74d831e4 tried to make the tests always use the latest Core version, but it caused nightlies to fail because we cannot install floating versions of packages because we have enabled [NuGet Central Package Management](https://learn.microsoft.com/en-us/nuget/consume-packages/central-package-management). This PR reverts that commit. The `.0` patch will still get used, and I left instructions in the tests for how to run tests with a subsequent patch version. Core patches still get tested in nightly builds, by building the Core from the `release-*` branch.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -144,7 +144,10 @@ jobs:
 
       - name: Run examples (.NET 5)
         shell: bash
-        run: find examples/ -name *.csproj | xargs -I{} dotnet run --project {} -f net5.0
+        # macos-latest runs on Apple Silicon, which .NET 5 does not support.
+        # The tests run fine for some reason, but the examples do not. Explicitly
+        # specify the architecture to avoid this issue.
+        run: find examples/ -name *.csproj | xargs -I{} dotnet run --project {} -f net5.0 -a x64
 
       - name: Run examples (.NET 6)
         shell: bash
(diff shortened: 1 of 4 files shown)


ihnorton (approved):

teo-tsirpanis:
Any objections in just skipping building the examples on macOS? There are complications with GitHub Actions switching `macos-latest` to Apple Silicon and .NET 5 not supporting it. If I build the examples just for .NET 6 they still are built for .NET 5 for some reason and preventing this is not trivial, with some early attempts of mine failing. I have no idea why the tests succeed at the same time.

The risk is low: the examples do not contain any platform-specific code and they will still run on Ubuntu and Windows.

Other things we can do:

* Stop targeting .NET 5 in the examples – the tests will still run in both .NET 5 and 6
* Roll back CI images `macos-12` – I prefer not to since it would introduce more technical debt in the future, and coverage on ARM64 is always a good thing.

@ihnorton thoughts?

ihnorton:
> Stop targeting .NET 5 in the examples – the tests will still run in both .NET 5 and 6

This seems preferable.

teo-tsirpanis:
CI is green. I have launched [a nightly build on my fork](https://github.com/teo-tsirpanis/TileDB-CSharp/actions/runs/8898152159). Once it succeeds, I will merge this and make a release.

teo-tsirpanis:
Nightlies are successful. Merging…

PR #425, opened 2024-05-03
Add VFS list example, with optional config

ihnorton:
Add VFS list example, with optional config for usage to set S3 or other parameters.

### DIFF
--- a/examples/TileDB.CSharp.Example/ExampleVFS.cs
+++ b/examples/TileDB.CSharp.Example/ExampleVFS.cs
@@ -1,0 +1,56 @@
+using System;
+using System.IO;
+using System.Collections.Generic;
+
+using TileDB.CSharp;
+using TileDB.CSharp.Examples;
+
+public class ExampleVFS
+{
+    private readonly string _array_path;
+    private readonly Context _ctx;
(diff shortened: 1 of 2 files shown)


Shelnutt2 (approved):

shaunrd0 (approved):

examples/TileDB.CSharp.Example/ExampleVFS.cs
```diff
@@ -0,0 +1,56 @@
...
+using TileDB.CSharp;
+using TileDB.CSharp.Examples;
+
+public class ExampleVFS
+{
+    private readonly string _array_path;
+    private readonly Context _ctx;
+    private readonly VFS _vfs;
+
+    public ExampleVFS()
```
teo-tsirpanis, line 14: I would prefer the class to be `static`, the fields to be local variables, the constructor to be merged with `Run`, and `CreateFiles` and `ListFiles` to be [local functions](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/local-functions) inside `Run`.

Feel free to make the adjustments or I will on Wednesday when I return.

PR #433, opened 2024-05-31
Update Core to 2.24.

teo-tsirpanis:


tests/TileDB.CSharp.Test/ArrayTest.cs
```diff
@@ -170,7 +170,43 @@ public void TestConsolidateFragments()
...
+
+        Assert.AreEqual(ObjectType.Invalid, context.GetObjectType(uri));
+    }
+
+    [TestMethod]
+    public void TestUpgradeVersion()
+    {
+        var context = Context.GetDefault();
+
+        using var uri = new TemporaryDirectory("array_upgrade_version");
```
shaunrd0, line 197: I don't think it will be upgraded :grimacing: aside from S3 or committing the array though I'm not sure how we could test it with an array from a previous format version. Any ideas?

https://github.com/TileDB-Inc/TileDB/blob/93220d7c3041fcd05fdbf8a45cc9969f47aaf763/tiledb/sm/array/array.cc#L2161

teo-tsirpanis: Indeed, this test only calls the function to check that nothing catastrophic happened. We could check in an empty array with an older format version (just the schema file would be tiny), but I am slightly to moderately against this.

The Python API is doing [something interesting](https://github.com/TileDB-Inc/TileDB-Py/blob/3ea00dbc2eabce972bea8af1ea8a4dc420e00a72/tiledb/tests/test_libtiledb.py#L311-L341). 👀

shaunrd0: Oh nice find! The format version is the first field in the array schema on disk.. Could we create an array and then adjust the first 4 bytes to set an older version?

https://github.com/TileDB-Inc/TileDB/blob/dev/format_spec/array_schema.md#array-schema-file

teo-tsirpanis: We can't do that easily because the schema is [compressed](https://github.com/TileDB-Inc/TileDB/blob/3ef79f748084d3969f641bfc9b02743caeb58913/tiledb/sm/tile/generic_tile_io.cc#L260-L263), and it would be very fragile either way.

teo-tsirpanis: Updated the test.

shaunrd0 (approved):

PR #450, opened 2024-07-10
Update Core to 2.25.

teo-tsirpanis:


teo-tsirpanis:
C# support for current domain will be added in a future release, once the C API provides facilities for type validation.

shaunrd0 (approved):

PR #487, opened 2024-11-16
Fix nightlies et al.

### LINKED ISSUE #485: Nightly GitHub Actions Build Fail on Sat, November 16th 2024
See run for more details:
https://github.com/TileDB-Inc/TileDB-CSharp/actions/runs/11866479269

teo-tsirpanis:
Fixes #485.

* The "Remove existing .NET versions" step was removed. It is not necessary because the .NET SDK version to use is pinned by the `global.json` file. ~~Moreover it's likely that the forced deletion of the `$DOTNET_ROOT` folder corrupts the installation and causes nightlies to fail because the `dotnet --info` command fails with an exception.~~
* The .NET SDK used was updated from 8.0.1xx to 8.0.4xx.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,3 +8,7 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    groups:
+      mstest:
+        # Update MSTest packages together, they are sometimes incompatible with each other.
+        patterns: ["MSTest.*"]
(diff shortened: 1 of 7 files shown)


teo-tsirpanis:
[Nightlies were fixed](https://github.com/teo-tsirpanis/TileDB-CSharp/actions/runs/11896557241) after my latest change.

shaunrd0 (approved):

PR #556, opened 2025-05-06
Clean-up and take advantage of newer .NET APIs.

teo-tsirpanis:
[SC-65510](https://app.shortcut.com/tiledb-inc/story/65510/take-advantage-of-newer-net-apis)

Now that we target .NET 8, this PR makes use of some APIs added after .NET 5, and cleans up some adjacent code.


sources/TileDB.CSharp/Marshalling/MarshaledContiguousStringCollection.cs
```diff
@@ -21,7 +21,7 @@ public MarshaledContiguousStringCollection(IReadOnlyCollection<string> strings,
         Encoding encoding = MarshaledString.GetEncoding(dataType);
         try
         {
-            Offsets = (ulong*)Marshal.AllocHGlobal(strings.Count * sizeof(ulong));
+            Offsets = (ulong*)NativeMemory.Alloc((nuint)strings.Count, sizeof(ulong));
```
teo-tsirpanis, line 24: For historical reasons, [`Marshal.AllocHGlobal`](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.marshal.allochglobal) uses a legacy and slower API on Windows to allocate memory. [`NativeMemory.Alloc`](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.nativememory.alloc?view=net-9.0) was introduced more recently and is documented to always call `malloc` while also featuring slightly improved ergonomics.

sources/TileDB.CSharp/Query.cs
```diff
@@ -965,60 +965,4 @@ private void CheckDataType<T>(string name)
     /// <exception cref="KeyNotFoundException">No validity buffer has
     /// been registered with <paramref name="name"/>.</exception>
     public ulong GetResultValidities(string name) => _handle.GetResultValidities(name);
-
-    private sealed class BufferHandle : IDisposable
```
shaunrd0, line 969: Nice catch!

shaunrd0 (approved):

PR #557, opened 2025-05-08
Update Core to 2.28.1.

teo-tsirpanis:


sources/TileDB.CSharp/Interop/Methods.cs
```diff
@@ -1333,6 +1333,10 @@ public static int tiledb_status([NativeTypeName("capi_return_t")] int x)
         [return: NativeTypeName("int32_t")]
         public static extern int tiledb_stats_disable();
 
+        [DllImport("tiledb", CallingConvention = CallingConvention.Cdecl, ExactSpelling = true)]
+        [return: NativeTypeName("int32_t")]
+        public static extern int tiledb_stats_is_enabled([NativeTypeName("uint8_t *")] byte* enabled);
+
```
teo-tsirpanis, line 1339: I was on the verge of not exposing it, as enabling and disabling global stats is meaningless, but since it's a stable API, I don't feel I have a choice.

shaunrd0 (approved):

